PR #58, opened 2026-05-04
fix: session namespace rollback, last_error for syntax/security, session_state noise, CLI help

pzfreo:
## Summary

Four bugs identified in code review, all fixed:

- **High**: Failed `execute()` leaked assigned variables into the namespace. A subsequent successful call could then promote stale geometry (e.g. a `result` set before a `raise`) to `current_shape`. Fixed by `_rollback_namespace()` which deletes any keys added during a failed execution, called in all failure branches.
- **Medium**: `last_error()` always returned `{"error": null}` after `SecurityError` or `SyntaxError` because those paths returned early before setting `last_error_detail`. Both paths now populate it before returning.
- **Medium**: `session_state().variables` included ~190 build123d symbols after `from build123d import *`, drowning out user variables. Added `_is_imported_symbol()` to filter modules and any class/callable whose `__module__` starts with `build123d`/`cadquery`/`OCP`. User Shape *instances* are not classes so they still appear.
- **Low**: CLI `--help` omitted `last_error`, `validate_code`, `shape_compare`, and `repair_hints` from the tool list.

## Test plan

- [ ] `uv run --python 3.12 pytest -q` → 198 passed (verified locally)
- [ ] After `execute("result = Box(1,1,1)\nraise ValueError('bad')")` followed by `execute("x = 1")`, `current_shape` is `None`
- [ ] After `execute("import os")`, `last_error()` returns `{"type": "SecurityError", ...}` not null
- [ ] After `execute("def nope(\n  pass")`, `last_error()` returns `{"type": "SyntaxError", ...}` not null
- [ ] After `execute("from build123d import *")`, `session_state()` variables section is empty (no build123d classes)
- [ ] `build123d-mcp --help` shows `last_error`, `validate_code`, `shape_compare`, `repair_hints`

### DIFF
--- a/src/build123d_mcp/server.py
+++ b/src/build123d_mcp/server.py
@@ -246,6 +246,10 @@ Available tools:
   save_snapshot     Save a named geometric checkpoint
   restore_snapshot  Restore geometry from a named checkpoint
   diff_snapshot     Compare two snapshots; format="json" for structured output
+  last_error        Details of the last failed execute() (type, message, line, excerpt)
+  validate_code     Check code for syntax/security errors before executing
+  shape_compare     Compare two named shapes by geometry metrics
+  repair_hints      Get fix suggestions for a given execute() error message
   version           Return the server version string
   workflow_hints    Return guidance on using these tools effectively
   reset             Clear the session (namespace, shapes, snapshots)
--- a/src/build123d_mcp/server.py
+++ b/src/build123d_mcp/server.py
@@ -154,7 +154,7 @@ def repair_hints(error_text: str) -> str:
 
 @mcp.tool()
 def last_error() -> str:
-    """Return details of the last failed execute() call: exception type, message, line number within the submitted code, and a 5-line excerpt around the failing line. Returns {\"error\": null} if the last execute() succeeded or no execute() has failed yet. Call this immediately after an execute() error to get the exact failing line — much faster than re-reading the submitted code."""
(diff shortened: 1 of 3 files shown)


pzfreo:
Three follow-up issues from review addressed in latest commit:

**1. High — full namespace value rollback**
`_rollback_namespace` now snapshots *values* (not just key names) before execution and restores them on failure. Changed `keys_before` (a set) to `values_before` (a dict of key→value). On rollback: new keys are deleted AND pre-existing keys are restored to their pre-execution values. The repro now works correctly:
```python
s.execute("from build123d import *\nresult = Box(10,10,10)")
s.execute("result = Box(1,1,1)\nraise ValueError('bad')")
s.execute("x = 1")
# current_shape is still the 10mm box ✓
```

**2. Low/Medium — SyntaxError now has an excerpt**
Added `_syntax_excerpt()` helper (same logic as `_make_error_detail`) called in the SyntaxError handler. Updated `last_error()` docstring to say excerpts are available for syntax and runtime errors, but not security blocks (no line number there).

**3. Low — `session_state().variables` now empty after `from build123d import *`**
Added a name-based filter using `dir(build123d)` to exclude any name that build123d exports, regardless of type. This catches constants (`MM`, `CM`), type aliases (`VectorLike`), and anything else build123d makes public. Result after `from build123d import *`: `variables: {}`.

PR #303, opened 2026-06-23
feat: recover() tool — heal an invalid solid, or fail without distorting it

pzfreo:
## Why

When `validate()` FAILs and the agent's hand-coded `ShapeFix`/sewing in `execute()` can't clear the defect — e.g. an **unorientable or un-meshable BSpline face** inherited from a malformed imported STEP — there is no programmatic heal in the server. `repair_hints` is static hint text, and nothing does defeaturing. So the agent burns turns and gives up, shipping an invalid solid that scores zero.

This surfaced on two CADGenBench editing fixtures whose **input STEPs are themselves BRepCheck-invalid**: the agent applied the edit correctly but couldn't clear the inherited face.

## What

A `recover()` tool that runs a heal ladder and keeps the first variant that passes the **exact** gate:

1. `ShapeFix_Shape` — cheap reorient/fix (what the agent already tries).
2. **Defeature** the BRepCheck-invalid faces — remove them, neighbours extend to close the gap.
3. **Drop + sew** the bad faces over a tolerance ladder, rebuild the solid, `ShapeFix_Solid`.

### The key safety property
A heal is accepted **only if it also leaves the volume and bounding box essentially unchanged (≤1% each)**. A heal that distorts the part is refused. So `recover()` **either returns a faithful valid solid or fails and leaves the geometry untouched** — it never hands back a mangled solid that merely happens to be watertight. (This was a deliberate design requirement: don't fool the caller into thinking it healed when it actually distorted.)

Requires a single solid; a multi-solid shape is refused rather than guessing which body to heal. A candidate is only sent through the expensive mesh gate once BRepCheck-clean, so the ladder stays cheap.

## Validation

- **Fixture A** (input STEP carries an unorientable BSpline face): defeature recovers it at **+0.005% volume / +0.584% bbox** — official CADGenBench gate PASSes the result.
- **Fixture B** (defect resists every geometry-preserving heal): correctly returns **FAIL, geometry untouched**.
- `tests/test_recover.py` (3 tests) pins the constructible contracts: already-valid no-op, single-solid guard, resolver error. The heal-success and single-solid fail-untouched paths are validated end-to-end against the fixtures above (an unorientable face can't be built reliably in a few lines of build123d). 42 repo tests pass (recover + worker + validate + server).

## Notes for review

- Wired as a `WorkerSession` op with the export-grade timeout budget (it runs exact gates).
- `_gate_report` calls are not individually try-wrapped (matches `validate()`'s assumption that the gate is robust) — flagging for discussion.


pzfreo:
## ⚠️ Shelved as draft — adversarial review found unresolved safety issues

Two rounds of adversarial review (3 reviewers each) repeatedly broke the core safety guarantee (*"never hands back a distorted solid that merely happens to be watertight"*). Documenting the blocking findings here for whoever picks this up.

### Round 1 (initial in-process design) → led to the out-of-process rewrite
- **CRITICAL: session-kill.** Defeature is an un-interruptible ~minutes-long native call; measured **307s on a 7MB fixture** vs a 120s op budget → worker SIGKILL → session lost. *(Addressed by moving the heal out-of-process, bounded.)*
- **HIGH: guard anchored to the invalid solid's signed volume** — false-rejected a faithful ShapeFix heal at "+50%" (reversed-face box: broken vol 666.7 vs true 1000). *(Addressed by switching to bbox guard.)*
- **HIGH: drop+sew could merge/fill thin real features and still pass.** *(Addressed by removing drop+sew.)*

### Round 2 (out-of-process rewrite, commit 16da2a8) — STILL BLOCKING
- **CRITICAL (reproduced): bbox-only guard is blind to internal-feature loss.** A 40mm block with a Ø16 through-bore → defeature removes the bore wall → solid block, **volume 55957 → 64000 (+14%), bbox delta 0.0 → accepted as a "faithful heal."** The round-1 distortion hole moved from the volume guard to the bbox guard; it did not close.
- **MAJOR: session-kill not fully closed.** The post-subprocess `import_step` of the healed STEP runs **in-process, unbudgeted, after** the bounded subprocess (`recover.py:156`). A heal that succeeds near the time limit + a slow re-import can still push the worker past its op-timeout → SIGKILL. `export.py` avoids this by re-importing *before* computing the remaining budget.
- **MEDIUM: narrow + untestable scope.** The ladder only acts on **BRepCheck-invalid faces**; the gate's primary target (mesh non-manifold edges/vertices) is BRepCheck-*valid*, so both rungs no-op on it. Constructible toy defects normalize on STEP round-trip, so the heal-success path has **no self-contained CI test** (only external benchmark fixtures 217/240).
- MEDIUM: 1-solid Compound + free faces/PMI silently dropped (`Solid(solids[0])`), unlike export which preserves them.
- MEDIUM: `"already valid"` after round-trip normalization isn't re-registered → contradicts `validate()` on the same object.

### Path forward if resumed
A genuinely safe guard likely needs a **surface-deviation (Hausdorff) bound** between original and healed, not cheap volume/bbox proxies — bigger work. Minimum to unblock: (1) volume guard between the two *reimported* solids (catches bore-fill), (2) bound the post-subprocess reimport, (3) document the BRepCheck-face-invalid scope, (4) a self-contained bbox-guard test via an injected over-moving heal.

**Validated wins that do hold:** 217 heals via defeature (0.64mm bbox, official gate PASS); 240's over-budget defeature is killed → FAIL untouched, session intact. The out-of-process bounding and fail-untouched-on-timeout contracts work.

pzfreo:
# Resume plan: proper Hausdorff fidelity guard + #308-style bounded timeout

This is the detailed plan for the two round-2 blockers (and the MEDIUMs), built on the now-established out-of-process precedent in #307 (policy) and #308 (render tessellation). It replaces the "Path forward if resumed" sketch above.

**The two halves map to the two blockers:**
- **Part A (Hausdorff)** closes the **CRITICAL** — bbox guard blind to internal-feature loss (Ø16 bore-fill: vol +14%, bbox Δ 0.0 → falsely accepted).
- **Part B (bounded reimport)** closes the **MAJOR** — post-subprocess `import_step` runs in-process/unbudgeted → can SIGKILL the session.

The key lesson from #308: its child returns a *cheaply-loadable* pickle of `(verts, tris)`, so **nothing heavy runs after the subprocess bound**. recover can't fully copy that — a healed BRep solid must go back through the OCC STEP reader to live in the session — so recover must instead *bound around* that reimport (Part B).

---

## Part A — Fidelity guard: symmetric surface Hausdorff

Replace the volume→bbox proxy chain (both already proven insufficient) with a **symmetric surface Hausdorff** bound between the original invalid solid and the healed candidate. Hausdorff is the actual invariant a "faithful heal" must satisfy: *no surface point moved more than ε.*

- bbox is blind to internal loss (the bore-fill). signed volume is blind to orientation defects (a reversed face halves it). Hausdorff sees both.
- **Worked example (the CRITICAL fixture):** original Ø16 bore-wall points sit ~12 mm from the nearest surface of the defeatured solid block → Hausdorff ≈ 12 mm → **reject**. The legit 217 sliver-defeature changes only a thin region → Hausdorff ≈ 0.6 mm → **accept**. Wide, safe separation (0.6 vs 12).

**Direction matters — must be symmetric.** `orig→healed` catches feature *loss* (the bore); `healed→orig` catches feature *addition*. Defeature only removes, but ShapeFix can do anything, so compute both and take the max.

**Computation — two tiers, OCC-only by default:**
1. **History-localized (fast path, defeature rung).** `BRepAlgoAPI_Defeaturing` exposes `Modified()/Generated()/IsDeleted()/History()`. Sample Hausdorff only on deleted/modified/generated faces + immediate neighbours — the rest is provably identical, so this is tiny and exact. (Confirmed available on the installed OCP.)
2. **Full bounded sampling (fallback, ShapeFix rung — no clean history).** Tessellate both solids fine (deflection ≤ ε/4); for each sampled point on A compute exact point-to-solid distance via OCC `BRepExtrema_DistShapeShape(vertex, solidB)` — **dependency-free and exact**. Cap sample count (~3000/direction). The whole computation runs inside the bounded subprocess, so if it overruns on a huge part it just FAILs-untouched — the #307 contract working *for* us: we can afford the correct expensive check precisely because it's bounded.

> `scipy` (cKDTree) would accelerate the fallback's candidate-triangle lookup, but it's currently only a **transitive** dep (1.17.1 present, not declared). The primary path uses OCC only, so no new hard dependency. If we adopt scipy, declare it explicitly in `pyproject.toml`.

**Tolerance.** `ε = max(absolute_floor, relative × bbox_diagonal)`, a documented module constant. Tune empirically against the fixtures: **217 must PASS** (~0.6 mm), **synthetic bore-fill must FAIL** (~12 mm), **240 stays FAIL**. Start at `ε = max(0.25 mm, 0.2 % diag)`. Emit the measured Hausdorff in `RECOVER_RESULT` so every accept/reject is auditable.

---

## Part B — #308-style bounded timeout (close the MAJOR)

- **Use export's *dynamic* budget, not #308's fixed 75 s.** A heal legitimately wants the whole remaining op budget, so: `remaining = max(60, exec_timeout) − elapsed − MARGIN`; spawn only if `remaining ≥ floor`. (#308's flat budget suits render because its bracket work is cheap; recover's isn't.)
- All heavy native work — ShapeFix, defeature, both reimport-gates, **and the Hausdorff** — runs in the subprocess under `subprocess.run(timeout=remaining)`. Mostly already there.
- **The actual fix for the unbudgeted parent reimport:** the child *already* reimports `out.step` to gate it. Have it **time that reimport** and emit `reimport_secs`. The parent then performs its own `import_step(out_step)` **only if** `time_left ≥ reimport_secs × SAFETY (≥2)`; otherwise it returns **FAIL-untouched** — and, rather than discard a successful heal, **persists `out.step` to a returned path** so the agent can `import_cad_file()` it on a fresh op. This is recover's analogue of "nothing heavy after the bound": we can't make the BRep reimport cheap, so we gate it on its own measured cost.

---

## Part C — fold in the round-2 MEDIUMs

- **Preserve non-solid content.** Stop `Solid(solids[0])` silently dropping a 1-solid Compound's free faces / PMI. v1: **refuse with a clear message** (simpler and honest) rather than guess-reattach; revisit preservation later.
- **Re-register on normalization.** When the reimported STEP normalises an in-memory-invalid solid to valid (`already_valid`), re-register the *reimported* solid so session `validate()` agrees with the artifact a scorer sees — guarded by the same Hausdorff bound, and reported as "normalised on STEP round-trip."
- **Document scope.** The ladder only clears BRepCheck-invalid *faces*; mesh non-manifold edges/vertices are BRepCheck-*valid*, so both rungs no-op on them. State this in the docstring + tool description; defer a non-manifold rung (sewing / ShapeUpgrade) to a follow-up.

---

## Part D — self-contained tests (close the "untestable" MEDIUM)

- **Unit-test Hausdorff directly with *valid* solids** (no need to construct an invalid one): block vs block-with-Ø16-bore → assert ≈ 12 mm; block vs block-with-0.3 mm-chamfer → assert ≤ ε; identical solids → 0. Pins the guard in isolation, in CI, no external fixtures.
- **Injected over-moving heal:** monkeypatch the defeature step to return a deliberately distorted solid; assert recover rejects it and leaves geometry untouched (round-2 "minimum to unblock" #4).
- **Budget-decline path:** force a tiny remaining budget; assert FAIL-untouched + persisted-path message + session intact.
- Keep the existing constructible contracts (already-valid no-op, single-solid guard, resolver error). Fixtures 217/240 remain the end-to-end check.

---

## Sequencing

1. **Hausdorff guard (A)** + unit tests (D) — replaces the bbox guard in `_recover_subprocess.py`.
2. **Bounded-reimport fix (B)** + budget-decline test.
3. **MEDIUMs (C).**
4. **Adversarial review** (3 reviewers, per the export/recover/#308 precedent) before un-drafting this PR.

## Risks / open questions

- **ε tuning is empirical.** The 0.6-vs-12 mm gap is comfortable on the known fixtures; confirm it holds across more parts before trusting the bound.
- **OCC point-to-solid cost** on large (~879-face) parts at fine sampling may FAIL-untouched on the biggest parts. The history-localized path (A.1) avoids this for the defeature rung; only the ShapeFix-rung fallback is exposed.
- **Persisting `out.step` on budget-decline** crosses recover's "in-session only" contract. Confirm that's wanted vs. a plain FAIL.

pzfreo:
## Resumed — both round-2 blockers closed, adversarial review re-run (commit b8a91dd)

Implemented the plan above against the **real cadgenbench fixtures** (fetched 217/240 via `harness/fetch_fixture.py`).

### CRITICAL (fidelity) — bbox guard → surface-deviation (Hausdorff)
Reproduced the hole first: a Ø16 bore-fill defeature is **volume 55957 → 64000 (+14%), bbox delta 0.0 → accepted** by the old guard. Replaced it with a symmetric surface-deviation bound — both surfaces sampled densely (triangulation nodes + triangle centroids), symmetric max nearest-neighbour distance via a `scipy.spatial.cKDTree`. Point-to-sample distance only ever **over**-states the true distance, so the guard never under-reports a distortion (one-sided-safe). ε = `max(1 mm, 1% diag)`.

Calibrated/validated on the fixtures:
- **217** (real invalid input): heals via defeature at **1.18 mm < its 1.96 mm tolerance** → PASS, ~17–20 s, well under budget.
- **Ø16 bore-fill**: reads **~16 mm ≫ ε** → refused.
- A pocket-fill reads ~4 mm → refused; identical solids → 0.

Two cost fixes were needed to make this real-time: the per-point OCC `BRepExtrema` approach was ~90 s on 217 → switched to cKDTree (~0.7 s); and the mesh is **cleaned before sampling** so density (and the reading) is deterministic, not contaminated by an earlier coarse gate mesh (that bug inflated 217 to a false 2.8 mm).

### MAJOR (session safety) — bounded parent reimport
The subprocess now **measures its own reimport cost** and reports it; the parent only performs its own (un-interruptible) `import_step` if `≥ 2×` that cost remains in the budget, else FAILs untouched and persists the heal to a path the agent can `import_cad_file()` on a fresh op. **240** (1004 faces, over-budget defeature) → killed at the bound, FAIL untouched, **session intact**, at both `exec_timeout` 120 s and 60 s.

### Adversarial review (3 reviewers) — and the fixes it forced
- **Fidelity reviewer**: could not produce a false PASS in ~50 constructed cases; confirmed the metric is mathematically one-sided (always over-reports). Only safe-direction notes (ε grows on large parts by design; planar sampling is over-conservative).
- **Session-safety reviewer**: could not break it; ~15 s worst-case margin, SIGKILL escalation confirmed, holds under CPU saturation.
- **Contract reviewer** found real bugs, now fixed:
  - **CRITICAL**: parent registered the reimport as PASS *without re-checking it*. Now the parent does a **BRepCheck-only** re-check (no meshing → cannot itself overrun the budget; the authoritative mesh gate stays in the bounded subprocess) that the load is a single valid solid — a corrupt/solid-less STEP can no longer be registered as PASS, and a missing `out.step` returns FAIL instead of raising.
  - **MAJOR**: `store_as` no longer hijacks `current_shape`.
  - **MEDIUM**: a 1-solid Compound carrying free faces/PMI is now refused, not silently stripped.

### Tests / status
793 pass. New tests pin: the surface-deviation guard (rejects bore-fill, accepts faithful), injected distorting/faithful heals through the subprocess, budget-decline, parent re-check, `store_as`, free-faces refusal. `recover` classified in the worker-boundary smoke inventory + added to the MCP tool-list test; `scipy` declared.

Left as documented limitations (not blocking): the planar-face sampling is coarse but over-conservative (safe direction); the `already_valid`-on-round-trip verdict still differs from in-memory `validate()` by design; budget-decline persists one STEP file per occurrence.

Still **draft** — over to you on whether to mark ready. Note the branch is behind `main` and will want a rebase before merge.

pzfreo:
## Review of the Hausdorff + bounded-reimport update

Checked the pushed branch against the two round-2 blockers and the points raised on the resume plan.

### ✅ Both blockers genuinely closed
- **CRITICAL (bbox-blind internal-feature loss).** The symmetric surface Hausdorff guard works: `res_calib.txt` shows the Ø16 bore-fill at `dev=16.3mm` vs `eps=1.0mm` → **rejected**, while fixture 217 (`diag=193 → eps=1.93mm`) passes. This is the right invariant and it catches what bbox/volume couldn't.
- **MAJOR (unbudgeted parent reimport → session-kill).** Closed — the parent now gates its `import_step` on the subprocess-measured reimport cost vs remaining budget (`_REIMPORT_SAFETY=2.0`), and persists `out.step` on budget-decline. The heavy work is all under the bounded subprocess.
- **ε floor** raised to **1% of the diagonal** (not 0.2%), so 217 passes — good catch on the earlier inconsistency.
- **scipy** now declared in `pyproject.toml` (it backs the `cKDTree` sampling). Good.

### ❌ Residual: Hausdorff is magnitude-only — small-feature blind spot remains
The guard rejects on Hausdorff magnitude alone; there's no topological check (no `IsDeleted`/`History`/`Modified` use). The calibration proves the gap is real: on a 193 mm part **ε = 1.93 mm**, so *any* heal moving surfaces < 1.93 mm is accepted as faithful — and **filling a Ø3–4 mm bore moves surfaces only ~1.5–2 mm → it would pass the guard.** The Ø16 case is caught only because it's large; a small internal-void wall being defeatured away slips under ε.

Suggested cheap insurance (defeature rung only): pair the magnitude bound with a topology check — *did defeaturing delete a face that bounded a closed internal void (a bore/pocket wall)?* `BRepAlgoAPI_Defeaturing` exposes `History()/IsDeleted()`, so this is exact and orthogonal to ε, and catches a small-bore-fill regardless of its tiny Hausdorff. Or: accept the ε-floor as the documented limit and state it explicitly in the docstring.

### 🧹 Housekeeping: stray scratch files
12 `res_*.txt` debug/calibration outputs are committed (`res_attack1.txt`, `res_calib.txt`, `res_dense.txt`, `res_underreport.txt`, …). Remove them before un-drafting and add `res_*.txt` to `.gitignore`.

### Net
The hard part is done well — the two blockers are properly closed. The only substantive open item is the small-feature window above (judgment call: add the void-bounding-face check vs document the ε limit), plus the file cleanup. Happy to run the full 3-reviewer adversarial round on this branch as the external check before un-drafting.

pzfreo:
## Addressed review + ran a second 3-reviewer adversarial round (commit 25f3349)

Thanks — both points from your review handled, and I cleaned up after myself.

### Housekeeping
The 12 `res_*.txt` were scratch output an adversarial-review agent wrote to the repo root that a `git add -A` swept in. Removed and `res_*.txt` is now gitignored.

### The small-feature concern — investigated hard, then both fixed and documented
I chased your suggestion empirically before deciding:

- **The topology check doesn't work here.** A *faithful* defeature changes the Euler characteristic too — χ shifts by **3** on fixture 217, and OCC's χ deltas aren't clean handle/void counts (a through-bore read −7, a blind pocket −29). So χ/shell/genus can't separate a faithful heal from a distorting one; it would false-reject the real fixtures. Disproven, with numbers.
- **The through-bore premise doesn't reproduce.** Filling a through-feature leaves its wall a full bore-width from any remaining surface, so the deviation is *large* regardless of diameter: Ø4 through-bore = **30 mm**, Ø5 near a thin wall = 100 mm. Caught, not a blind spot. (A second independent reviewer confirmed this across many constructions.)
- **You were right about counterbores, though** — that's the real window. A counterbore/spotface over a *retained* through-hole moves only ≈√(annular-gap² + depth²) ≈ **~6 mm**, *independent of part size*, because the hole wall is the nearby surface. With ε = 1% of the diagonal, a ≥~620 mm part has ε > 6 mm → a standard cap-screw counterbore would read as faithful. **Fixed by capping ε at 3 mm** (`min(max(1 mm, 1% diag), 3 mm)`): refuses the counterbore/spotface cases (5.4–6.3 mm) while still clearing fixture 217 (ε 1.93 mm, dev 1.18 mm). The residual — a feature moving every surface point < ε — is irreducible for a pure surface-distance metric and is now documented accurately (not the under-stated "shallow pocket a few tenths deep").
- **Doc bug fixed:** the docstrings said "0.5% of the diagonal" while the constant is 1% — corrected to the capped formula in both places + CHANGELOG.

### Second adversarial round (3 reviewers) — net clean
- **Fidelity false-PASS hunt:** could not get a >ε distortion accepted in ~dozens of constructions; `dev ≥ ground-truth` in every case (one-sided-safe confirmed vs exact OCC point-to-face). STEP round-trip preserves geometry to ~1e-13 mm, so "measured on cand, gated on reimp" can't be opened.
- **Session safety / contract:** clean. The new parent BRepCheck re-check is **bounded** (0.5 s on the 1004-face healed 240 solid, ~0.3× the import cost it follows) and runs after the budget gate without risking the op-timeout — worst margin **14.93 s**. FAIL-untouched holds on every path; 240 over-budget → killed, session intact at exec_timeout 60 and 120.
- **Contract fix verified:** the parent re-check rejects corrupt / solid-less / multi-solid / missing `out.step` (FAIL untouched, never registered as PASS). Also fixed a `current_shape` quirk a reviewer found (an in-place `store_as == object_name` was leaving `current_shape` pointing at the old instance).

### Remaining documented limitations (not blocking)
- A sub-ε feature can still be lost (irreducible for this metric class) — now documented honestly.
- The parent re-check re-validates *validity* but trusts the subprocess's *fidelity* verdict; unreachable in production (the subprocess ships the exact `cand` it measured), a documented trust boundary.
- Heal coverage is the targeted class: a single solid with BRepCheck-invalid faces. Open-shell (0-solid) and BRepCheck-valid-but-mesh-broken inputs no-op — the common ones an agent hits; worth a follow-up but out of scope here.

Tests green (full suite). Still draft — your call on ready/merge, and on whether the 3 mm cap is the right trade-off (it errs toward false-reject = fail-untouched, matching the "refuse rather than distort" philosophy; tighten/loosen as you see fit).

pzfreo:
## Closing — not shipping. Post-mortem so this isn't re-attempted the same way.

The two round-2 blockers got closed (Hausdorff fidelity guard, #308-style bounded reimport) and two rounds of adversarial review couldn't break the safety contract. But a corpus-level check of the fidelity guard — the whole reason `recover()` exists — shows it can't be validated, so the auto-accept gate shouldn't ship.

### The evidence: the fidelity threshold is calibrated on n = 1
I healed **all 32 CADGenBench editing fixtures** (`harness/fetch_fixture.py`), not just 217/240:

- **29 of 32 are already valid** — no BRepCheck-invalid faces, so recover's defeature rung never fires.
- **3 are invalid** (202, 217, 240) — the only candidates recover can act on.
- **Only 217 produces a successful, in-budget heal.**
  - **202** (diag 632 mm): defeature runs but does **not clear** the invalid face → FAIL.
  - **240** (1004 faces): defeature is **over budget** → FAIL (timeout).

So ε (`min(max(1 mm, 1% diag), 3 mm)`) — and the 1% term and 3 mm cap specifically — rest on a **single** successful heal (217 @ 1.18 mm). The corpus offers no second data point to validate against.

### The two populations overlap exactly where it matters
| | surface deviation |
|---|---|
| the one faithful heal (217) | **1.18 mm** |
| real feature removals that actually changed geometry (204/224/231) | **1.6 / 3.5 / 3.7 mm** |

Every gross distortion is caught — but the faithful heal and the smallest real distortions sit right next to each other. There's no valley to put a threshold in. That's the inherent limit: a surface-distance metric **cannot distinguish "healed a 1.5 mm defect" from "removed a 1.5 mm feature"** — both move the surface by the same amount. No amount of tuning ε fixes that; ε is a cut through a continuum where the classes overlap.

### Why that kills the auto-accept gate
`recover()`'s promise is "returns a faithful valid solid or fails untouched." An auto-accept gate calibrated on n = 1, sitting on overlapping populations, gives the agent **false confidence on exactly the cases it gets wrong**. That's worse than not having the tool — a silent distortion shipped as PASS scores zero and the agent trusts it. Combined with the narrow reach (fires on 3/32, heals 1/32), the cost/benefit isn't there.

### What was genuinely solid (and reusable)
- **Out-of-process, op-budget-bounded heal** with measured-reimport guard so it can never SIGKILL the worker (#307 policy, same pattern as export's mesh gate and #308's render tessellation). Two reviews confirmed ~15 s margin and FAIL-untouched on the over-budget 240. This pattern stands on its own, independent of recover.
- The adversarial-review-driven contract fixes (parent re-check, store_as/current_shape, free-faces refusal) were correct — they just don't rescue the premise.

### For anyone reconsidering
A trustworthy "heal without distorting" guard needs ground-truth-aware fidelity (does the heal stay faithful to *intended* geometry), not a self-referential surface-distance threshold — that's a much bigger problem than a tolerance. Absent that, the honest shape of this tool is **advisory**: heal, report the measured deviation, and let a human/agent decide — not an auto-accept gate. Given recover heals 1/32 fixtures, even that may not be worth it.

Closing as won't-ship. The session-safety/timeout pattern is the keeper.

Issue #311, opened 2026-06-24
Investigate render_view reliability under load (flaky render-in-process test may signal a production race)


pzfreo:
## Concern

`tests/test_in_process_session.py::test_stateful_tool_works_in_process[render_view]` is **intermittently failing**, and the worry is not test hygiene — it may indicate that **`render_view` itself occasionally fails under load in production**, which is the exact condition a **parallel benchmark sweep** creates (many MCP sessions saturating one machine). A spurious render failure is costly: it wastes an agent's turn, or worse, the agent draws a wrong conclusion from a missing image.

## Evidence so far

- The test **passes 3/3 in isolation** but **failed once inside a 14-minute full-suite run** (800 passed / 1 failed).
- That full-suite run was **fixed-order** (`-p no:randomly`), so this is **not** a pytest-randomly ordering artifact.
- It is **always the same test** (`[render_view]`), never other in-process tools.
- It also wobbled once during the #308 work — so it has now recurred, which is why it warrants real root-causing rather than a re-run.
- The failure surfaces only as a bare `assert False` (the smoke check "did render produce a PNG"); **the underlying reason has not yet been captured.**

## Why it's plausibly a real production race (not just a test artifact)

Since #308, `render_view`'s path is:
1. write a STEP per shape, then `subprocess.run` a tessellation worker (bounded), then
2. **VTK offscreen render** to PNG (in-worker / daemon path).

Both stages are contention-sensitive: under a saturated machine the tessellation subprocess start can be slow, and **VTK offscreen rendering** is a classic source of load-dependent flakiness (shared GL/offscreen state, buffer allocation). Either could intermittently yield no PNG / an error return — for a real agent as well as the test.

## What to determine

1. **Capture the actual failure mode.** Instrument so a failed render reports *why*: tessellation `TimeoutExpired`? subprocess non-zero exit / empty pickle? VTK returned no image? An empty/garbage PNG vs no file?
2. **Contention vs ordering.** Re-run the fixed-order full suite a few times: deterministic repro at the same spot → an ordering/state-leak interaction; non-deterministic → load/timing contention.
3. **Reproduce the production mechanism directly** (independent of the test harness): drive the real `render_view` code path repeatedly under induced CPU contention (N busy background procs) and measure the failure rate + mode.
[...]

Issue #143, opened 2026-06-01
Windows + Codex MCP host: execute fails with "Worker process failed to start within timeout", while WorkerSession works in PowerShell


huazhi-sd:
## Summary

I am testing `build123d-mcp` on a Windows workstation through the Codex MCP host.

The server starts and simple tools work:

- `version` returns `0.3.28`
- `reset` returns `Session reset.`

However, any call to `execute()` fails, even a trivial print-only script.

## Environment

- OS: Windows
- Python: 3.12.10
- build123d-mcp: 0.3.28
- MCP host: Codex desktop / Codex MCP
- Launch mode tested:
  1. `uv tool run --python 3.12 build123d-mcp --exec-timeout 30`
  2. local launcher with `if __name__ == "__main__": main()`
  3. fixed local venv + local launcher

## Failing behavior

Calling:

```python
print("hello")
x = 2
print(x)
```

through the MCP `execute` tool returns:

```text
Error: Worker process failed to start within timeout.
```

After that, worker-dependent tools such as `session_state` also fail with the same worker startup issue.

## Important comparison

The same package and environment work when calling `WorkerSession` directly from normal PowerShell.

Example local script:

```python
from build123d_mcp.worker import WorkerSession

if __name__ == "__main__":
    session = WorkerSession(exec_timeout=30)
    print("version:", session.version())
    print(session.execute("print('hello')\nx=2\nprint(x)"))
    print(session.session_state())
```

This succeeds in PowerShell and returns the expected output.

A simple geometry test also succeeds in PowerShell:

- creates a `40 x 20 x 3 mm` plate
- cuts two `3.4 mm` holes
- measures correct bbox
- exports STEP successfully

## What I suspect

This may be related to Windows `multiprocessing.spawn` when `build123d-mcp` is launched as a stdio MCP server by Codex.

The issue does not seem to be build123d itself or `Session.execute()`, because direct `WorkerSession` execution works outside the MCP host.

## Full notes

I documented the trial here:

https://github.com/huazhi-sd/hardware-cad-agent-lab/blob/main/logs/ai-cad-project-watch/build123d-mcp-windows-codex-trial-2026-06-01.zh-CN.md

Happy to test a patch or provide more logs if useful.

huazhi-sd:
## Follow-up: no-worker diagnostic launcher works

I did one more diagnostic test to separate the MCP stdio host from the `WorkerSession` / Windows multiprocessing layer.

[...]

PR #313, opened 2026-06-24
[codex] Add localized surface deviation compare

pzfreo:
## Summary

- Adds bounded localized surface-deviation output to `shape_compare` while preserving the existing volume, bbox, topology, and center delta fields.
- Runs tessellation and nearest-neighbour surface comparison in a hard-bounded subprocess, with in-process fallback when child process creation is blocked.
- Adds explicit `scipy` runtime dependency for `cKDTree` use.
- Updates the MCP tool description to frame this as model-vs-input edit verification, not a score.
- Adds outcome tests for no-op comparisons, localized edits, unrelated-region changes, subprocess fallback, and timeout handling.

## Validation

- `uv run pytest tests/test_outcomes.py -q` -> 60 passed
- `uv run pytest tests/test_worker_timeouts.py tests/test_render_tessellation_bounded.py tests/test_locate.py -q` -> 23 passed
- `uv run ruff check src/build123d_mcp/_shape_compare_subprocess.py src/build123d_mcp/tools/shape_compare.py src/build123d_mcp/server.py tests/test_outcomes.py`
- `uv run ruff format --check src/build123d_mcp/_shape_compare_subprocess.py src/build123d_mcp/tools/shape_compare.py src/build123d_mcp/server.py tests/test_outcomes.py`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ dependencies = [
     "build123d-drafting-helpers>=0.10.0",
     "defusedxml>=0.7.1",
     "augura>=0.1.5",
+    "scipy",
 ]
 
 [project.optional-dependencies]
--- a/src/build123d_mcp/_shape_compare_subprocess.py
+++ b/src/build123d_mcp/_shape_compare_subprocess.py
@@ -144,6 +144,10 @@ def _regions(
 
 
 _EXACT_VOL_TOL = 1.0  # mm^3 — ignore boolean slivers below this
+# Below this displacement (mm) the change is a cut/flush-fill with ~0 surface
+# movement (the new surface sits where material was) — report volume, not a
+# misleading 0 displacement.
+_DISP_NULL_MM = 0.05
 
 
 def _chunk_displacement(chunk: Any, ref_solid: Any) -> float:
@@ -401,7 +405,21 @@ def compare_shapes(
             "~0 added/removed volume — likely tessellation noise the detector over-fired on, or a "
             "purely tangential rearrangement; treat as effectively unchanged."
         )
-    return _result(exact["displacement"], "exact_boolean", exact)
+        return _result(exact["displacement"], "exact_boolean", exact)
+
+    magnitude = exact["displacement"]
+    if magnitude < _DISP_NULL_MM:
+        # A cut / flush fill (drilled hole, blind pocket): the new surface lies where
+        # the removed material was, so the exact vertex displacement is ~0 even though
+        # volume changed. Surface displacement is the wrong magnitude here — reporting
+        # max_deviation=0 reads as "no change". Use the mesh estimate for the headline;
+        # the exact magnitude is the added/removed VOLUME.
+        magnitude = primary["max_deviation"]
+        warnings.append(
+            "this edit removes/adds material with ~0 net surface displacement (a cut or flush fill) — "
+            "max_deviation shown is the mesh estimate; the EXACT magnitude is removed_volume/added_volume."
+        )
+    return _result(magnitude, "exact_boolean", exact)
 
 
 def main(a_step: str, b_step: str, out_json: str, eps: str, budget_s: str = "0") -> None:
--- a/src/build123d_mcp/_shape_compare_subprocess.py
+++ b/src/build123d_mcp/_shape_compare_subprocess.py
@@ -30,6 +30,13 @@ from typing import Any
 # of the op budget, skip it and report the (flagged) mesh estimate instead of risking
 # a hard subprocess kill that would lose the whole result.
 _BOOL_RESERVE_S = 25.0
+# Skip the exact boolean when the changed region's clip box spans more than this
+# fraction of the part diagonal AND the part is large enough for that to be slow:
+# boolean cost scales with clip-box size, so a spread or large-extent edit on a big
+# part overruns the op budget. Such edits fall back to the (flagged) mesh estimate.
+# Small parts are fast regardless of spread, so the gate ignores them.
+_MAX_CLIP_DIAG_FRAC = 0.35
+_MIN_DIAG_FOR_CLIP_GATE = 300.0
 
 # eps (the per-vertex "this point moved" threshold) = factor x mesh deflection,
 # placed above the independent-tessellation noise floor. Deflection is diag*1e-3
@@ -226,7 +233,11 @@ def _exact_region_magnitude(
 
 
 def compare_shapes(
-    shape_a: Any, shape_b: Any, eps: float = 0.0, deadline: float | None = None
+    shape_a: Any,
+    shape_b: Any,
+    eps: float = 0.0,
+    deadline: float | None = None,
+    on_mesh_ready: Any = None,
 ) -> dict:
     """Symmetric vertex-sampled surface diff, localized to significant changed regions,
     with an EXACT boolean magnitude in each region.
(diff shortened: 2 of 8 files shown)


pzfreo:
## Updated per the adversarial review (commit 81d0e10)

The review found the localized output **untrustworthy on real imported STEPs**: a fixed `eps=0.01mm` sat an order of magnitude below the independent-tessellation noise floor, so a same-geometry re-export no-op fabricated a **5.49mm "change"** on fixture 248, and the 204 hole-move read identical to its own no-op. All findings addressed and **validated empirically on the real v0.3.57 fixture pairs**:

**Fixes**
- **eps auto-scales to the mesh deflection** (`3×deflection`, floor 0.05mm) instead of a fixed mm; both shapes tessellate at a **shared deflection**. (`eps>0` still overrides, for tests.)
- **Headline `max_deviation` = the largest *real* change** (over significant regions), `0` when nothing moves above the floor — a no-op now reads ~0, not the noise tail. `raw_surface_max` keeps the unfiltered value.
- **Regions filtered by spatial *span* (≥2.5×deflection), not vertex count** — density-independent, so it drops scattered noise specks on dense parts and still catches features on sparse ones. `changed` is the **dominant** region (no more centroid in empty space); per-region centroid/bbox in `regions[]`.
- **`unchanged_elsewhere` = region centroids clustered within 0.4×diag** — scale-independent (fixes both the fragmented-ring false-positive and the small-part misfire).
- **Tangential-move warning** when no region is found (a slid hole reads ~0 surface deviation — the metric's blind spot), echoed in the tool `note`.
- **New real-STEP round-trip no-op regression test** — the old identical-`Box()` test couldn't catch this (two identical boxes tessellate byte-identically).

**Before → after (measured)**

| case | before | after |
|---|---|---|
| 248 re-export no-op | max_dev **5.49**, `unchanged=false` | max_dev **0.0**, 0 regions, `unchanged=true` |
| 248 edit (+10mm raise) | 10.0 (noise-mixed) | **10.0**, region **at the block**, `unchanged=true` |
| 208 edit (fillet r3) | `unchanged=false` (ring→arcs) | `unchanged=true`, region at the rim |
| 204 edit (hole move) | "barely changed, localized" (wrong) | 0 regions + **tangential-move warning** |
| 243 edit (messy thicken) | — | `unchanged=false`, 1019 regions (honestly widespread) |

All 7 `shape_compare` tests pass (incl. the new round-trip no-op); ruff + mypy clean. Subprocess bounding / OSError fallback / framing were already solid and unchanged.

Worth a re-review of the new eps/region logic before un-drafting.

pzfreo:
## Bounded the exact boolean per the deep review (commit 19125f6)

The review confirmed the exact magnitude is **correct** (sign convention, displacement with no inflation, deterministic — all verified) but found a HIGH safety bug: the boolean overran the op budget on **~27% of real parts (4/15)** and the whole result was discarded on the hard kill (`_BOOL_RESERVE_S` only gated *starting* it; clipping to the union bbox of all regions blew up on spread edits). Fixed:

- **Clip-size gate** — skip the exact boolean (→ flagged mesh estimate) when the changed region spans >35% of the part diagonal **and** the part is large (>300mm), since boolean cost scales with clip-box size. Validated on the timed-out fixtures: **211 (was >105s timeout) → 13s mesh_estimate; 247 → 35s; localized 248 still exact** (add=3589, disp=10).
- **Salvage on timeout** — the worker persists the mesh-estimate result *before* the boolean (`on_mesh_ready`); the driver returns that flagged result on `TimeoutExpired` instead of discarding everything. So an overrun degrades to a usable estimate, never nothing.
- **In-process fallback** now passes a soft deadline (can't run the boolean unbounded on a subprocess-blocking host).
- **Reconciliation warning** when the mesh detector fires many regions but the boolean nets ~0 volume (the 238 noise/tangential case).

Tests added: pure-remove (drill hole → removed>0/added=0), budget-skip→mesh_estimate, large-spread→skip. **11 shape_compare tests pass; ruff+mypy clean.** Full `test_outcomes` had only the unrelated #311 render-under-load flake (passes 2/2 isolated).

pzfreo:
## Fixed the focused-review findings (commit c507e6a)

The focused review confirmed the salvage mechanism is sound but caught a real hole in the gate's thresholds:

- **CRITICAL — gate hole:** the `_MIN_DIAG_FOR_CLIP_GATE=300` part-size floor let **206 (diag 291mm) run the boolean 360s** (and hang in-process). **Fixed:** gate on **absolute clip size (>150mm)** — boolean cost scales with absolute, not relative, clip size. Validated: 206 → **23s mesh_estimate** (was 360s), 229 → 4s, tight-clip 217 → still exact (75s), localized 248 → still exact.
- **Hang on in-process path:** no op-timeout there to bound a runaway boolean. **Fixed:** in-process runs **mesh-only** (`allow_exact=False`).
- **Exception discards salvage:** post-boolean `displacement`/`volume` calls were outside the try, so a raise overwrote the persisted mesh salvage with an error. **Fixed:** extended the guard — a crash now falls back to the salvage, same as a timeout.

Tests added: wide-clip skip on a <300mm part (the 206 class), in-process mesh-only, exception-after-boolean→mesh, and the driver `TimeoutExpired` salvage branch. **14 shape_compare tests pass; ruff+mypy clean.**

PASS (unchanged): salvage-on-timeout end-to-end, finally/except ordering, corrupt-file handling, reconciliation warning.

pzfreo:
Review findings from the latest PR head (`c507e6a`):

1. **Critical: `shape_compare` can still kill the worker on slow geometry.**

   `shape_compare` computes its internal subprocess timeout from `session.exec_timeout` in `src/build123d_mcp/tools/shape_compare.py`, so with the default 120s exec timeout it gives the child roughly 105s (`exec_timeout - 15s margin`). But `WorkerSession.shape_compare()` is still routed with the fixed `_GEOMETRY_TIMEOUT` of 60s in `src/build123d_mcp/worker.py`.

   That means a large compare can block in `subprocess.run(timeout=~105)` while the parent kills the worker at 60s, losing session state and potentially leaving the child process running. This violates the main acceptance requirement that the tessellation/compare be subprocess-bounded below the worker op timeout.

   Suggested fix: give `shape_compare` its own worker budget function, probably `max(_GEOMETRY_TIMEOUT, ws._exec_timeout)`, and update the worker timeout coverage so the parent-side budget and internal subprocess budget cannot diverge.

2. **High: exact boolean displacement reports `max_deviation=0.0` for real cut/removal edits.**

   `_chunk_displacement()` measures only boolean-difference vertices against the reference shape. For a through-hole removal, the removed cylinder’s vertices lie on the remaining boundary/top-bottom faces, so vertex-to-reference distance is zero even though the surface changed substantially.

   Repro on PR head:

   ```python
   base = Box(60, 30, 10)
   show(base, 'a')
   show(base - Cylinder(4, 10), 'b')
   shape_compare('a', 'b')
   ```

   Result includes `magnitude_method='exact_boolean'`, `removed_volume=502.65`, and `raw_surface_max=31.8599`, but top-level `max_deviation=0.0` / `changed.local_max_deviation=0.0`. That is actively misleading for cut/remove operations. The new removal test checks volume only, so it misses this regression.

   Suggested fix: either do not replace the surface-deviation headline with this vertex-only exact displacement, or compute displacement from representative surface samples/faces rather than only vertices. At minimum, add a test asserting the removal case returns a non-zero displacement or falls back to the mesh estimate instead of reporting exact zero.

Validation I ran locally on PR head:

- `uv run pytest tests/test_outcomes.py -q -k shape_compare` -> 14 passed
- Manual repro above exposed the zero-displacement cut/removal issue.

I would request changes before merge; the timeout mismatch is the release blocker.

pzfreo:
Both findings were spot-on and are fixed in `36d4f5f`.

**1. Worker-op timeout mismatch (the release blocker) — fixed.** You're right: the driver bounded the subprocess by `~exec_timeout-15s` while the `shape_compare` worker op was still `_GEOMETRY_TIMEOUT=60`, so a >60s compare let the parent kill the worker mid-flight. Routed `shape_compare` through `_export_budget` (`max(60, exec_timeout)`) — identical to the subprocess budget formula and the same pattern `export`/`locate` already use, so the child is always bounded *below* the worker op timeout. Updated `test_worker_timeouts`: removed `shape_compare` from the fixed-geometry-timeout group and added `honours_exec_timeout` (300→300) + `keeps_export_floor` (10→60) so the parent and subprocess budgets can't diverge again.

**2. Cut/removal `max_deviation=0.0` — fixed.** Agreed it's actively misleading. `_chunk_displacement` is correct for adds/shaves but a drilled hole's removed-chunk vertices lie *on* B's new surface → distance 0. Now: when the exact displacement is ~0 but volume changed (a cut / flush fill), the headline `max_deviation` falls back to the (non-zero) mesh estimate with a warning that `removed_volume`/`added_volume` is the exact magnitude — so a cut never reads "no change." The removal test now asserts `max_deviation>0` + the warning (you were right that the volume-only assertion missed it).

`uv run pytest -k shape_compare` → 14 passed; `tests/test_worker_timeouts.py tests/test_worker_boundary_coverage.py` → 40 passed; ruff + mypy clean.

pzfreo:
Rereviewed latest PR head (`36d4f5f`). The previous two blockers are mostly addressed, but I found one remaining API/semantics issue before merge:

**High: `magnitude_method` is misleading for cut / flush-fill cases.**

For cuts where exact displacement is near zero but volume changed, `_shape_compare_subprocess.py` now replaces `max_deviation` with the mesh estimate, but still returns `magnitude_method='exact_boolean'`.

That conflicts with the public note in `shape_compare.py`, which says `max_deviation` is exact when `magnitude_method='exact_boolean'`. Machine consumers will reasonably trust that field as exact, even though in this branch it is a mesh-estimated fallback for these cases.

Repro on latest head:

```python
base = Box(60, 30, 10)
show(base, 'a')
show(base - Cylinder(4, 10), 'b')
shape_compare('a', 'b')
```

Result now correctly avoids the old `max_deviation=0.0` problem and reports exact removed volume, but it returns:

```json
{
  "max_deviation": 31.8599,
  "magnitude_method": "exact_boolean",
  "changed": {
    "removed_volume": 502.65,
    "added_volume": 0.0
  },
  "warnings": ["... max_deviation shown is the mesh estimate ..."]
}
```

The warning is accurate, but the structured method field is not. This matters because agents/tools will key off `magnitude_method`, not parse warning prose.

Suggested fix: return a distinct method such as `mesh_estimate_with_exact_volume`, or set `magnitude_method='mesh_estimate'` while still including exact `added_volume` / `removed_volume` fields. Then update the removal test to assert the method semantics, not just `max_deviation > 0`.

Resolved from prior review:

- Worker timeout mismatch is fixed: `shape_compare` now uses `_export_budget` and has timeout coverage.
- The cut/removal zero-displacement case no longer reports `max_deviation=0.0`; it now reports a non-zero mesh estimate plus exact removed volume.

Validation I ran locally on latest PR head:

- `uv run pytest tests/test_outcomes.py tests/test_worker_timeouts.py -q -k 'shape_compare or worker_timeout or timeout'` -> 27 passed
- Manual through-hole repro above exposed the remaining `magnitude_method` inconsistency.

pzfreo:
Fixed in `cfbc7b0` — you're right that the structured field has to be self-consistent for machine consumers.

The cut / flush-fill case now returns **`magnitude_method='exact_volume_mesh_displacement'`** (not `exact_boolean`): the added/removed **volumes are exact**, but `max_deviation` is a mesh estimate (a cut has ~0 true surface displacement). The tool `note` now documents all three values — `exact_boolean` (displacement + volumes exact), `exact_volume_mesh_displacement` (volumes exact, max_deviation mesh-estimated), `mesh_estimate` (both mesh) — and states volumes are exact whenever the method starts with `exact_`. The removal test asserts the method (`exact_volume_mesh_displacement`), not just `max_deviation>0`.

14 shape_compare tests pass; ruff + mypy clean.

pzfreo:
Fresh adversarial review on current PR head (`cfbc7b0`):

**High: in-process fallback can still crash `shape_compare` on tessellation failures.**

The normal subprocess path has a structured error boundary: `_shape_compare_subprocess.main()` catches exceptions and writes JSON. But the `OSError` fallback path in `src/build123d_mcp/tools/shape_compare.py` calls `_surface_compare_in_process()` directly, and that helper directly returns `compare_shapes(...)`.

Inside `compare_shapes`, `_tessellate_points()` calls `shape.tessellate(deflection)` without catching exceptions. So on a subprocess-blocked host (#143 / in-process mode), an OCC/build123d tessellation quirk like:

```text
AttributeError: 'NoneType' object has no attribute 'NbNodes'
```

propagates out of `shape_compare()` instead of returning a JSON `surface_deviation.error` / graceful compare result.

I reproduced this locally on PR head by forcing `subprocess.run` to raise `PermissionError` and one shape’s `tessellate()` to raise the `NbNodes` AttributeError; `shape_compare()` raised instead of returning JSON.

Suggested fix: wrap the in-process fallback in the same structured-error boundary as the subprocess worker, e.g. catch exceptions in `_surface_compare_in_process()` and return `{"error": "...", "warnings": [...]}` rather than letting them escape. Add a regression test for `subprocess.run -> OSError` plus a tessellation exception.

Validation I ran:

- `uv run pytest tests/test_outcomes.py tests/test_worker_timeouts.py -q -k 'shape_compare or worker_timeout or timeout'` -> 27 passed
- `uv run ruff check src/build123d_mcp/_shape_compare_subprocess.py src/build123d_mcp/tools/shape_compare.py src/build123d_mcp/worker.py tests/test_outcomes.py tests/test_worker_timeouts.py` -> passed
- Manual adversarial repro exposed the fallback crash above.

pzfreo:
Fixed in `3bbb95c` — correct, the in-process fallback didn't share the subprocess worker's structured-error boundary. `_surface_compare_in_process` now wraps `compare_shapes` in the same try/except, returning `{"error": ..., "warnings": []}` instead of letting a tessellation exception (the 0.11 'NbNodes' quirk) escape `shape_compare()`. Added a regression test: `subprocess.run`→OSError + a tessellation `AttributeError` → the tool returns a JSON surface error, doesn't raise. 15 shape_compare tests pass; ruff + mypy clean.

PR #316, opened 2026-06-25
Add an opt-in live session viewer over a Unix domain socket

provos:
I was really glad to find build123d-mcp. It solves many of the problems I had
using build123d with an agent before. This is a small PR, and I mainly wanted to
see whether you are open to changes like this. If the approach works for you, I
am happy to add more functionality on top of it.

## What and why
build123d-mcp gives an agent fixed-viewpoint `render_view` PNGs. For work that
needs human spatial or aesthetic judgment, a still is a poor channel: you want to
grab the model and spin it as it changes. This PR adds an opt-in live viewer.
Start the server with `--viewer-socket PATH` and it streams the session's
geometry to a connected viewer over a Unix domain socket, updating after each
geometry-mutating tool call. The agent still gets its PNGs while a human watches
the model evolve.

This PR is the server-side publisher (the socket interface) only. A native viewer
window is left as a follow-up so the change stays small and free of new
dependencies. A minimal example consumer using only the standard library is
included.

## Why it is safe to merge
- **Inert by default.** With no `--viewer-socket`, nothing changes. When the flag
  is set but no viewer has attached, the agent path pays nothing (it skips
  tessellation and the worker round-trip entirely).
- **Never blocks the agent.** The publisher runs on a background daemon thread.
  Bounded, drop-oldest per-client buffers throttle a slow viewer rather than the
  producer.
- **No new dependencies.** It reuses the existing bounded out-of-process
  tessellation and encodes to glTF-binary (glb) with a small self-contained
  writer, so it adds no third-party package and keeps OCC and VTK out of the
  encode path.
- **Respects the worker boundary.** Only identity-changed shapes are tessellated,
  in the worker process where the geometry lives, and only vertex and triangle
  arrays cross the pipe. A worker restart emits `RESET` so clients drop a stale
  scene.
- **Outside the sandbox.** The socket lives in the server process, not the
  sandboxed `execute()` worker, so user code cannot reach it.
- POSIX only (AF_UNIX), validated and gated in the CLI.

## Protocol
Length-prefixed frames: `[u32 json_len][json][u32 bin_len][payload]`. Events are
`HELLO`, `UPSERT` (with a glb payload), `REMOVE`, and `RESET`. A viewer that
attaches mid-session receives `HELLO` followed by a full-scene dump. The full
spec is in `docs/live-viewer.md`.

## Testing
The full suite passes (`uv run pytest`, 826 passed), and ruff and mypy are clean.
New tests cover the glb encoder, the wire framing, publisher behaviour, and an
end-to-end path from `server.execute` through the worker to the socket.

## Not in scope, possible follow-ups
A graphical viewer (pyvista, or a browser three.js page over a UDS to WebSocket
bridge), session persistence and replay, and a viewer to server channel for
tweaking parameters from the viewer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.3.60
+
+### Added
+
+- **Live session viewer over a Unix domain socket (`--viewer-socket PATH`).** Optionally stream the session's geometry to an interactive 3D viewer so a human can watch and rotate the model while an agent drives the MCP, instead of only seeing fixed-viewpoint `render_view` PNGs. When the flag (or `BUILD123D_VIEWER_SOCKET`) is set, the server binds a UDS and, after each geometry-mutating tool (`execute`, `reset`, `restore_snapshot`, `import_cad_file`, `load_part`), broadcasts the **changed** shapes as glTF-binary (glb) to connected viewer clients. The publisher runs on a background daemon thread in the server process and never blocks the agent path: it pays nothing while no viewer is attached (no tessellation, no worker round-trip), tessellates only identity-changed shapes via the existing hard-bounded out-of-process path, and a slow client is throttled (bounded per-client buffers, drop-oldest) rather than the producer. A viewer attaching mid-session gets `HELLO` + a full-scene dump; a worker restart emits `RESET`. The glb encoder is a small self-contained glTF 2.0 writer, so there is **no new dependency** and no OCC/VTK in the encode step. The socket is created mode 0600 and lives outside the sandboxed `execute()` worker. Wire protocol, usage, and a dependency-free reference consumer (`examples/live_viewer_client.py`) are documented in `docs/live-viewer.md`. POSIX only.
+
 ## v0.3.59
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -320,6 +320,21 @@ Clients connect to `http://<host>:8000/mcp`.
 
 > **Security note.** The server has no built-in authentication. When binding to `0.0.0.0`, place it behind a reverse proxy with auth (e.g. nginx + mTLS, or a VPN).
 
+### Live session viewer (experimental)
+
+Start the server with `--viewer-socket PATH` (or `BUILD123D_VIEWER_SOCKET=PATH`) to
+stream the session's geometry to an interactive 3D viewer over a Unix domain socket,
+so a human can watch and rotate the model while an agent drives the tools. Each
+shape is broadcast as a glTF-binary (glb) and updated after every geometry-mutating
+tool call. The publisher runs on a background thread in the server process and does
+not stall the agent path; a pure agent run (no `--viewer-socket`) does no extra work.
+
+Two example consumers ship with it, neither a package dependency: an interactive
+pyvista window ([`examples/live_viewer_pyvista.py`](examples/live_viewer_pyvista.py),
+run via `uv run --with pyvista`) and a dependency-free text client
+([`examples/live_viewer_client.py`](examples/live_viewer_client.py)). The wire
+protocol and design are in [docs/live-viewer.md](docs/live-viewer.md). POSIX only.
+
 ---
 
 ## System prompt
--- a/docs/live-viewer.md
+++ b/docs/live-viewer.md
@@ -1,0 +1,98 @@
+# Live session viewer
+
+Stream the build123d session's geometry to an interactive 3D viewer over a Unix
+domain socket (UDS), so a human can watch and rotate the model while an agent
+drives the MCP server. The agent keeps getting its fixed-viewpoint `render_view`
+PNGs; the human gets a live view that updates as the session changes.
+
+This page documents the **server-side publisher** (the UDS interface). A native
+viewer window is a separate, optional consumer. A minimal stdlib example ships in
+[`examples/live_viewer_client.py`](../examples/live_viewer_client.py), and a richer
+graphical viewer can be built on top of the same protocol.
+
+## Enabling it
+
+Start the server with a socket path:
+
+```bash
+build123d-mcp --viewer-socket /tmp/b123d.sock
+# or
+BUILD123D_VIEWER_SOCKET=/tmp/b123d.sock build123d-mcp
+```
+
+POSIX only (uses `AF_UNIX`). When the flag is absent the feature is fully inert
+and a pure agent run pays nothing. When the flag is set, the server keeps a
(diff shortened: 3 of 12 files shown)


pzfreo (commented):
First off — thank you for this. It's a lovely PR, and it genuinely made my day to see someone engage with the internals this thoughtfully. You clearly read the architecture closely, and it shows everywhere.

Full disclosure on how I'm working here: I ran your PR through Claude for a proper adversarial review — so the detailed feedback below is "vibe-coded," Claude's analysis rather than me hand-auditing every line. I've read it and I agree with it, but I want to be upfront about where it came from. It was also very complimentary about your code, which matches my own first impression. 🙂

**What it loved** (and so do I): the worker boundary is respected — tessellation stays in the worker, only vertex/triangle arrays cross the pipe; you reuse the bounded, hard-killable out-of-process tessellation path so it can't trip our op-timeout/SIGKILL; the crash-restart `RESET` is wired correctly; the HTTP per-tenant guard keeps tenant geometry out of the shared stream; the glb encoder is spec-correct; and the drop-oldest backlog logic never corrupts a partially-sent frame. Plus real tests and the boundary-coverage classification. That's a lot to nail on a first pass.

One bigger-picture thought before the nitpicks, since you asked whether we're open to changes like this — yes, very much. The only thing I want to be honest about is that it's a real scope addition: build123d-mcp is agent-facing today (fixed-viewpoint PNGs), and this adds a human-in-the-loop *watching* channel, which I think is a great direction. My one hesitation about merging the publisher on its own is that it commits us to maintaining a wire protocol + glb encoder for something nobody can see yet. So if you're up for it, I'd love to land this alongside even a minimal consumer — a small pyvista window, or a UDS→WebSocket bridge to a three.js page — so it ships as something visible. Totally fine as a fast follow if you'd rather keep this PR tight; just want to decide it together.

Two things Claude flagged that I'd want to tidy before merge, both because they touch claims the PR makes:

1. **`_flush` holds the lock across its whole send loop.** A fast local viewer draining a big backlog can make the publisher thread hold `self._lock` for a burst, and the agent path's `upsert()` (called after every mutating tool) has to acquire that same lock — so the agent *can* be briefly blocked on contention. It's local-UDS-fast and bounded, so in practice tiny, but it means "never blocks the agent" is really "rarely, briefly blocks." Either releasing the lock per-frame in `_flush`, or just softening the wording, would square it.

2. **Bind→chmod TOCTOU on the socket.** `bind()` then `chmod(0o600)` leaves a brief window where the socket has umask-default perms, so the "restricted to 0600 right after bind" claim has a small gap. Low severity (microsecond, local-only, just geometry), but an `os.umask(0o077)` around the bind closes it cleanly.

And a few genuine nice-to-haves, no pressure on any of them:

3. The server-side `_cache` holds an encoded glb per named shape on top of the worker's copy — effectively a second copy of the geometry. Fine in practice; maybe just a note in `docs/live-viewer.md`.
4. `start()` unlinks whatever path it's given without an `S_ISSOCK` check — a tiny guard would avoid nuking a real file if someone passes the wrong path.
5. A late attach with >256 named shapes could drop HELLO/early UPSERTs from the full-scene dump (the cap is on frame count). Unrealistic for CAD, but worth a comment.

Honestly though — this is great work, and none of the above changes that. Thank you for offering to build more on top of it; I'd be genuinely glad to have you. Let's figure out the direction + the visible-consumer bit together and take it from there.

provos:
Thanks for the thoughtful reply, and for being open to new features. I appreciate it.

All the concerns from your review are addressed, and I have added a live viewer to this PR so the publisher ships as something you can actually see. Two consumers come with it, neither a package dependency: an interactive pyvista window and a dependency-free text client.

Here is how to try the window (two terminals, POSIX):

**1. The viewer** (pulls in pyvista just for this run):

```bash
uv run --with pyvista python examples/live_viewer_pyvista.py /tmp/b123d.sock
```

**2. Drive a quick build.** Save this as `demo.py` and run `uv run python demo.py`:

```python
import time
from build123d_mcp import server
from build123d_mcp.worker import WorkerSession

STEPS = [
    "from build123d import *",
    "show(Box(40, 24, 4), 'frame')",
    "show(Box(40, 24, 4) - Pos(-12,0,0)*Cylinder(radius=7,height=10)"
    " - Pos(12,0,0)*Cylinder(radius=7,height=10), 'frame')",
    "show(Pos(-12,0,0)*Cylinder(radius=6.5,height=2), 'lens_left')",
    "show(Pos(12,0,0)*Cylinder(radius=6.5,height=2), 'lens_right')",
]

if __name__ == "__main__":
    server.configure(WorkerSession(exec_timeout=60))
    server.start_viewer("/tmp/b123d.sock")
    for code in STEPS:
        print(server.execute(code).splitlines()[0])
        time.sleep(3)
    input("rotate the model, then press Enter to quit\n")
```

Grab the model and rotate it as the frame builds, the lens holes get cut, and the lenses drop in. Starting the viewer after the build finishes exercises the late-attach path, and restarting the server shows the viewer reconnect on its own.

For the real workflow, point build123d-mcp at `--viewer-socket /tmp/b123d.sock` in your MCP client and drive an agent with the viewer open. The wire protocol and design are in `docs/live-viewer.md`.

pzfreo:
Thanks for the update! It's looking good. I'll test it out tomorrow (its late here) and let you know how it works for me.

pzfreo:
Thank you, @provos — this is a genuinely well-built contribution. The design is conservative where it counts: the publisher lives outside the sandbox, the socket is owner-only (`umask(0o077)` across `bind()` + `chmod 0600`, no creation window), inbound bytes are ignored so there's no parser to attack, and it only ever streams the session's own geometry. The HTTP multi-tenant guard in `_publish_deltas()` and the worker-restart RESET hook show you thought through the edges. CI is green across the full matrix and it merges clean.

Two small **non-blocking** hardening notes for a possible follow-up (I'll open an issue):

1. **Per-client buffer isn't byte-bounded.** `_MAX_QUEUED_FRAMES = 256` caps frame *count*, and the on-connect full-scene dump uses `cap=False`, so a stalled local viewer can drive server RSS up with large meshes. Note this buffer lives in the server process, which `--memory-limit-mb` doesn't bound (that limits the worker). A byte budget on the deque would close it.
2. **Socket placement.** Worth a docs nudge toward `$XDG_RUNTIME_DIR` over `/tmp`, since a world-writable dir lets another local user pre-create a live socket and make startup refuse (same-user nuisance, not a compromise — `start()` correctly won't clobber a file or steal a live socket).

Neither holds up the merge. Really nice work — thanks for the careful PR and the example clients.

PR #345, opened 2026-07-01
fix: subprocess tools never surface a cold-start null (field report #2)

### LINKED ISSUE #344: Cold-start null on first out-of-process subprocess tool call (field report #2)
Field report from the CADGenBench pipeline: the **first** invocation of a subprocess-backed tool in a fresh worker session returned `null` **~100%** of the time (30/30 first-calls across `locate_gate_defects` @0.3.59 and `design_audit` @0.3.60); later calls in the same session mostly succeed. Deterministic cold-start, not flakiness.

**Root cause (confirmed in code):** the first out-of-process spawn pays a cold `from build123d import *` (OCP is hundreds of MB of `.so`) in a fresh child and blows its budget on call #1 only; and — unlike `execute()` — the `@_op` tool proxies let the resulting worker-kill/op-timeout `RuntimeError` propagate, which the MCP client renders as a bare `null` (the tool functions themselves always return JSON). Affects `locate_gate_defects`, `design_audit`, and by shared machinery `export`/`render_view`/`shape_compare`.

**Fixed in this PR:**
- **A (guaranteed):** the `@_op` proxy catches `RuntimeError`/`ExecutionTimeout` → returns a structured `{"error": …}` result (like `execute()`), so the client never sees `null` and the true cause is visible in transcripts.
- **B (mitigation):** a daemon thread at worker startup primes `build123d` in a fresh child, warming the OS page cache so no user-facing first call is cold.

**Residual / needs your data:** ~20% of *later* calls still null (field report #1's truncation finding — likely genuine budget on heavy parts); triage separately. And because I can't reproduce the cold container locally, please confirm on a cold run that (a) the first-call `null` is gone / replaced by a real result, and (b) if any error remains, its now-visible `{"error": …}` message pins the exact mechanism.

Source: cadgenbench-build123d field report #2. Fixed by #NNN.

pzfreo:
Closes #344. Acts on field report #2: the **first** call of a subprocess-backed tool in a fresh worker session returned `null` **~100%** of the time in a cold container (30/30 across `locate_gate_defects` and `design_audit`); later calls succeeded.

## Root cause (confirmed in code)
The first out-of-process spawn pays a cold `from build123d import *` (OCP is huge) and blows its budget on call #1; and — unlike `execute()` — the `@_op` tool proxies let the resulting worker-kill / op-timeout `RuntimeError` propagate, which the MCP client renders as a bare `null`. (The tool functions themselves always return JSON; the `null` is the proxy raise surfacing as an MCP error.)

## Fix A — never surface `null` *(guaranteed, tested locally)*
The `@_op` proxy now catches `RuntimeError`/`ExecutionTimeout` and returns a structured `{"error": …}` result, exactly as `execute()` already does. The client always gets a readable, distinguishable reason instead of `null` — and this **makes the true cause visible in transcripts**, so it doubles as the diagnostic for the cold container I can't reproduce on a warm dev box. The low-level `_call` state-loss-on-timeout **raise** contract is unchanged (tests assert that directly on `_call`).

## Fix B — warm the cold-start *(mitigation)*
A daemon thread at worker startup primes `build123d` in a fresh child, warming the OS page cache so no user-facing first call is cold. Best-effort, off the critical path, swallowed on hosts that block subprocess creation (#143).

## Honest caveat
I **could not reproduce** the null locally — my disk cache is warm, so first `locate` returns in 5.5 s vs ~4.7 s warm. So **Fix B's efficacy needs validation on a cold CADGenBench run**; Fix A is the guaranteed improvement (null → readable error) regardless, and will reveal the exact mechanism if any error remains. The residual ~20% *later*-call null (field report #1 truncation) is triaged separately.

## Tests & docs
Proxy converts worker-kill/timeout → JSON error (new test); two tests that asserted the old raise (boundary unknown-object, in-process contract) updated to the structured-return surface; `export`/`render`/`_call` raise paths verified unaffected. Docs: CHANGELOG (v0.3.61). **867 passed**; ruff + mypy + `ruff format --check` clean.


pzfreo:
Closing — investigation against the actual run transcripts shows **there is no cold-start null bug** (see #344 / #346). The `null` was the streaming `status=in_progress` placeholder, not a tool result. Fix A (never-raise→structured) would also have *hidden* genuine `isError` failures, and Fix B (warmup) targets the wrong cost (per-process Python import, which disk-cache warming doesn't remove). Superseded by #346 (invariant test).

PR #356, opened 2026-07-04
Write single-solid STEP exports as one product, not a one-component assembly

faisal-shah:
## What

A single-solid `export("part.step")` is written as a one-component STEP *assembly* instead of a part. In a CAD kernel (SolidWorks, Inventor) it opens as an assembly with the body nested under a component and offset from the part origin, and in a stock document it can import blank. Web/mesh viewers show it fine because they flatten product structure, so it is easy to miss.

The file comes out shaped like this:

    PRODUCT('COMPOUND') -> PRODUCT('<part>')  +  NEXT_ASSEMBLY_USAGE_OCCURRENCE

## Why

It falls out of the `_write_step` workaround for gumyr/build123d#1356. On build123d 0.11 `export_step` raises `RuntimeError: Failed to write STEP file` on a solid that came straight from `import_step` -- which is most parts once they have round-tripped through `import_cad_file` (heavy builds, editing an imported STEP, assembling from part files). The workaround wraps the solid in a `Compound` to get back through the CAF writer, but a `Compound` with one child is written as an assembly, so every part that hits the fallback lands as a one-component assembly.

It shipped green because the validity gate only checks geometry (watertight / manifold / solid count), and a Compound-wrapped solid still re-imports as one valid solid. Nothing looked at product structure.

## Fix

#1356 is about the build123d `Solid` wrapper, not the geometry: a fresh `Solid` over the same `TopoDS` gets the CAF writer to accept it. Measured against build123d 0.11.1:

| fallback | writes? | product structure | body name |
|---|---|---|---|
| `Compound(children=[solid])` (old) | yes | `COMPOUND` -> child, **NAUO=1** | kept |
| raw `STEPControl_Writer` | yes | single product | **lost** |
| **`Solid(solid.wrapped)` + retry** | yes | **single product, NAUO=0** | **kept** |

So for a single solid, `_write_step` reconstructs and retries the CAF path -- one product, name/colour intact. The `Compound` wrap stays for genuine multi-solid exports, where the assembly structure is correct; the raw writer stays as the last resort. The gate also warns when a single-solid export still carries `NEXT_ASSEMBLY_USAGE_OCCURRENCE`, to catch a regression before it ships.

This is the mcp-side fix for the #1356 side effect; the underlying `export_step` regression is still open upstream. The `Solid(wrapped)` result is a useful data point for it -- a fresh wrapper over identical geometry fixes the write, which points at the XCAF registration path, not the geometry.

## Tests

`test_export_single_solid_is_not_written_as_assembly` imports a solid (the real #1356 trigger on 0.11), exports it, and asserts the file is a single product (no `NEXT_ASSEMBLY_USAGE_OCCURRENCE`) with the name preserved -- holds on 0.10 (primary path) and 0.11 (reconstruct). Existing export tests are unchanged and green. Full local suite: 888 passed, 22 skipped (one unrelated Windows symlink-privilege test).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.3.65
+
+### Fixed
+
+- **`export()` writes a single STEP part as one product, not a one-component assembly.** The build123d 0.11 `export_step` regression (gumyr/build123d#1356) raises on an import-derived solid, and the previous fallback wrapped it in a `Compound` to get through the CAF writer. But a `Compound` with one child is written as an assembly — `PRODUCT('COMPOUND')` -> child + `NEXT_ASSEMBLY_USAGE_OCCURRENCE` — so a CAD kernel (SolidWorks, Inventor) opens the file as a one-component assembly, with the body off the part origin under a nested component, importing blank in a stock document. Mesh viewers hid it by flattening product structure, and the validity gate only checked geometry, so it shipped green. `_write_step` now reconstructs a single solid (`Solid(shape.wrapped)`) and retries the CAF path — which #1356 accepts — keeping the file a single product with the body name/colour intact; the `Compound` wrap is reserved for genuine multi-solid exports, where the assembly structure is correct. The gate also warns when a single-solid export still carries `NEXT_ASSEMBLY_USAGE_OCCURRENCE`, catching a regression before it ships.
+
 ## v0.3.64
 
 ### Added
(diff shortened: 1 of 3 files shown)


pzfreo:
Thanks for the excellent PR. Much appreciated

faisal-shah:
Very welcome, this project is much appreciated.

pzfreo:
Thanks for the feedback!

PR #379, opened 2026-07-05
docs: clarify validate/export gate behavior

### LINKED ISSUE #320: docs: clarify validate() (fast screen) vs export() (exact authoritative gate) + add field-tested user gotchas
## Context

A documentation audit (driven by heavy benchmark use) found one wording error and several undocumented, field-tested gotchas in the user-facing docs (`skills/b123d-modeling`, `skills/b123d-drawing`, `default_prompt.md`, `llms.md`, `README.md`).

## 1. Clarify the intended `validate()` vs `export()` asymmetry (the one error)

The docs describe `export()` as "re-runs this gate" — implying export and validate run the **same** check. They don't, **by design**: `validate()` is the fast in-loop validity screen (called ~8.7×/fixture in our runs — it deliberately can't pay the exact cost every time), and `export()` runs the slower exact mesh-stitch check and is the authoritative verdict. The current wording leads a user to trust a `validate()` PASS and be surprised when `export()` rejects it (coincident faces, near-tangent junctions).

> Note: this is the intended *fast-screen vs exact-gate* asymmetry. It is distinct from #290, which is about `export()` being too *lenient* relative to the official CADGenBench gate.

Proposed wording: *"`validate()` is a fast validity screen; `export()` runs a slower, stricter exact check and is the authoritative verdict. Expect `export()` to occasionally reject geometry `validate()` passed — test-export to a throwaway path before finalizing."*

**Open question for the maintainer to document:** is a `validate()`-PASS-then-`export()`-FAIL an accepted by-design outcome (the fast screen is *allowed* false-PASSes), or a bug worth reporting? The docs should state which so users know how to react.

## 2. Add these field-tested gotchas (currently undocumented)

- **`is_inside()` over a point grid** is slow on large solids and can hit the op-timeout (which can lose session state — #307). Use `cross_sections()` / `render_view(clip_plane=...)` to inspect interiors instead.
- **`render_view` (high-quality / clipped) on a big import** can hit the timeout and wipe session state (#307); inspect numerically when the part is large.
- **`find_holes` on curved/BSpline sheets:** measure by the bore **axis**, not face/bounding-box centres (off-axis → wrong centre-to-centre). "Already at target" is a red flag to **re-measure**, not a task already done. (`find_holes` already returns the bore axis — surface this in the docs.)
- **Coincident / same-domain faces don't fuse** — the universal additive-edit footgun (~33 of our field notes): interpenetrate, bury the base 1–2 mm, or extend-and-trim with one planar cut, rather than butting faces exactly.
- **Heal decision note:** prefer `ShapeFix_Solid` over `ShapeFix_Shape` (the latter can globally re-orient and collapse volume); run heavy booleans as a standalone script when they exceed the worker timeout, then re-import.

## Framing

Keep this guidance **general CAD wisdom** — the benchmark-specific scoring framing (absolute scale, topology-as-face-graph) should NOT leak into the user docs. Our ~88 accumulated field notes are a ready source to mine further if useful.

Related: #307, #290, #286, #309.

ahfoysal:
## Summary

Fixes #320.

This clarifies that `validate()` is the fast in-loop screen while `export()` runs the stricter authoritative pre-export gate. It also adds a few practical modeling gotchas around point-grid `is_inside()`, clipped renders on large imports, curved-face hole axes, coincident additive faces, and heavy import repair.

## Checks

- `git diff --check`
- Tried `uv run pytest tests/test_validate.py -q`, but `uv` is not installed in this local shell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,10 @@ When using an AI to write build123d scripts, the AI writes blind — it cannot s
 
 **Import / export**
 - `export` — export as STEP / STL / DXF / SVG (or comma-separated like `step,stl`); auto-detects 2D vs 3D shape and routes to the appropriate format; targets a named object, the current shape, or `*` for all objects as an assembly
+- `validate` — fast pre-export validity screen. `export` repeats the gate with a
+  stricter, larger-budget check and is the authoritative verdict, so test-export
+  important parts before finalizing if the geometry has coincident faces,
+  near-tangent joins, or large imported B-reps
 - `import_cad_file` — load a STEP or STL file as a named object for comparison
 
 **Comparison**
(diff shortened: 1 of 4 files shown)


pzfreo:
Thanks for the well thought out contribution.

PR #380, opened 2026-07-05
Fix design audit baseline timing

### LINKED ISSUE #343: design_audit: baseline-relative per-perturbation cap is inflated by one-time build123d import (loose budget fairness)
Minor follow-up from the adversarial review of the design_audit budget-fairness fix (the O1 observation).

## What
The per-perturbation cap scales with the measured baseline rebuild time:

```python
# _design_audit_subprocess.run_audit
t_base = time.monotonic()
state["baseline"] = evaluate_program(program, cap(per_run_cap))   # <- first call
base_time = time.monotonic() - t_base
perturb_limit = max(_MIN_PERTURB_S, min(per_run_cap, int(base_time * _BASELINE_CAP_FACTOR) + 2))
```

The **first** `evaluate_program` call (the baseline) is the one that triggers the heavy one-time `from build123d import *` / OCP import (~2–4 s) inside the fresh subprocess. So `base_time ≈ import + build`, and `perturb_limit = K × base_time` is inflated by the import cost — a pathological perturbation is bounded to (say) ~18 s instead of the ~4 s the "4× baseline" intends.

## Impact
Low. It's **safe** (errs generous — never false-fails a slow-but-valid rebuild) and still a large improvement over the pre-fix `remaining()`-only cap that let one rebuild eat the whole budget. But the fairness guarantee is looser than nominal: one slow/pathological perturbation can still consume more of the budget than intended, starving later params sooner.

## Fix options
- **Warm the import once before timing the baseline** (e.g. a trivial `evaluate_program` warm-up, or import build123d in the subprocess `main()` before `run_audit`), so `base_time` reflects true rebuild time and `perturb_limit` is tight.
- Or subtract an estimated import constant from `base_time` before scaling.

## Scope note
Interacts with the deferred **high-leverage-first ordering** idea in #341 — if params are reordered/prioritised, per-perturbation fairness matters less. Could be folded in there. Filing separately so the specific `base_time`-includes-import nuance isn't lost.

ahfoysal:
Fixes #343.

Warms the build123d import before timing the baseline rebuild, so the per-perturbation cap is based on rebuild time instead of one-time import cost.

Tested:
- PYTHONPATH=src .venv/bin/python -m pytest tests/test_design_audit_subprocess.py
- .venv/bin/python -m ruff check src/build123d_mcp/_design_audit_subprocess.py tests/test_design_audit_subprocess.py
- .venv/bin/python -m ruff format --check src/build123d_mcp/_design_audit_subprocess.py tests/test_design_audit_subprocess.py

### DIFF
--- a/src/build123d_mcp/_design_audit_subprocess.py
+++ b/src/build123d_mcp/_design_audit_subprocess.py
@@ -32,6 +32,11 @@ _BASELINE_CAP_FACTOR = 4
 _MIN_PERTURB_S = 8
 
 
+def _warm_build123d_import() -> None:
+    """Load build123d before timing the baseline rebuild."""
+    __import__("build123d")
+
+
 def _const_number(node):
     """Return the numeric value of a constant / signed-constant node, else None.
 
@@ -284,6 +289,8 @@ def run_audit(
         with open(tmp, "w") as f:
             json.dump(state, f)
         os.replace(tmp, out_path)
+
+    _warm_build123d_import()
 
     # Baseline gets the full per-run budget (it may be the heaviest single build);
     # time it so perturbations can be capped relative to it.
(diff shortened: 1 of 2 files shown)


pzfreo:
Thanks for the contribution. Reviews clean and ready to merge

Issue #382, opened 2026-07-06
recover() follow-up: face-accounting fidelity (not ε), the field-proven heal ladder, and a report-not-verdict contract (#303 post-mortem revisited)


pzfreo:
## Revisiting recover() (#303): each leg of the won't-ship verdict has since changed

#303 closed won't-ship on three findings: the Hausdorff fidelity guard sat on overlapping
populations with no valley for a threshold, the threshold was calibrated on n = 1
successful heal, and reach was 1/32 fixtures. All correct at the time. Field evidence and
design lessons accumulated since then change each leg — this issue proposes a redesigned
follow-up, NOT a reopen of the old contract.

## 1. The guard measured the wrong thing — magnitude instead of locality

The #303 post-mortem's kill shot was: *"a surface-distance metric cannot distinguish
'healed a 1.5 mm defect' from 'removed a 1.5 mm feature' — both move the surface by the
same amount."* True — but surface distance was never the right discriminator.

In round 1 a topology check via `BRepAlgoAPI_Defeaturing.History()` was suggested, and
dismissed after testing **Euler-characteristic deltas** (χ shifted by 3 on the faithful
217 heal, −7/−29 on bores/pockets — genuinely useless). But χ was a proxy for the wrong
invariant. `History().IsDeleted()` / `.Modified()` gives exact **face accounting**: which
input faces did the heal delete or modify? The correct fidelity invariant is not "did the
surface move less than ε" but:

> **the changed-face set must be confined to the defective faces and their edge-adjacent
> ring.**

The bore-fill case — the failure that broke the ε-guard twice in adversarial review —
deletes *every healthy face of the bore wall*, faces far outside the defect's
neighbourhood. Under face accounting that is a set-inclusion violation: exact,
combinatorial, **threshold-free**. No ε means no calibration, which dissolves the n = 1
problem entirely. The "1.5 mm heal vs 1.5 mm feature removal" continuum stops being
ambiguous because a feature removal touches many healthy faces and a faithful heal
touches almost none.

(A residual judgment case exists: a *small feature whose own faces are the defect* — e.g.
a counterbore whose wall is BRepCheck-invalid. Removing it passes face accounting because
the deletion set is within one ring of the defect. That case is genuinely undecidable
without design intent, which is exactly why the tool must report rather than verdict —
see §3.)

## 2. Reach is not 1/32 — the field logs show 3/3, with a longer ladder

[...]

Issue #381, opened 2026-07-06
validate()'s fast-fallback mesh check is structurally blind to open edges — misleads the agent mid-session on large shapes


pzfreo:
## `validate()`'s fast-fallback mesh check is structurally blind to open edges — misleads the agent mid-session on large shapes

Follow-up to #360/#364 (closed/merged): that work correctly fixed the *crash* — `validate()`
no longer kills the worker on a large solid. This issue is the *accuracy* problem #364's fix
left behind: on a large/slow-to-stitch shape, `validate()` now degrades **gracefully** to the
fast coordinate-weld check instead of crashing, but that fast check is not a coarser version
of the exact one — it is **hardcoded to report `open_edges = 0`** regardless of the true
value (`validate.py` — the fallback branch sets `mesh_open_edges = mesh_untri_faces = mesh_nmv
= 0` and only ever populates `mesh_nm_edges` from `_mesh_defects`, which by design checks
non-manifold edges only, never open ones). So on a shape big enough to blow the exact check's
35s wall-clock budget (`_GATE_MESH_BUDGET_S`), every `validate()` call in the session reports
`PASS` even when the shape has real open edges — not a probabilistic miss, a guaranteed one.

## Field evidence

Fresh from a real 81-fixture CADGenBench sweep (`gpt55-v0368-xhigh-full`, mcp 0.3.68 — i.e.
**after** #364, so this is the residual accuracy gap, not the crash recurring):

- **Fixture 240** (editing): every one of its 15 `validate()` PASS results carried
  `"mesh_check": "fast"`. Every subsequent `export()` to the real output target failed with
  the identical defect the whole session — `"4 mesh open edge(s) — the tessellated boundary
  is not closed"` — across every repair strategy the agent tried (a healed rebuild, a
  single-patch rebuild, a re-exported healed check). The agent's own verification loop
  (`execute() → validate() → …`, which the prompt explicitly tells it to trust) reported a
  clean shape the entire time; only `export()` (which runs the mesh check out-of-process,
  unbounded by the 35s budget) ever saw the truth, and by then there was no signal pointing
  at *which* step introduced the defect.
- We independently hit the identical mechanism in `package_submission.py`'s packaging-time
  proxy check (fixture 202, a different run) and fixed it there by routing through
  `_run_mesh_gate_subprocess` — the same out-of-process path `export()` already uses — with a
  generous one-off timeout. That fix is a good existence proof that the pattern works; it
[...]

PR #383, opened 2026-07-06
feat: b123d-repair skill — a validity-gate repair cookbook

pzfreo:
## Why

The server can *diagnose* an invalid solid — `validate()` names the defect class, `locate_gate_defects()` gives its coordinates — but offers no guidance on *fixing* one. `repair_hints` covers Python/API errors, not geometry. Field evidence from the latest full CADGenBench sweep: fixture 217 (BRepCheck-broken imported STEP) burned its **entire session** on failed heal attempts — 22 straight `validate()` FAILs — and shipped invalid, scoring zero. Every recipe that would have healed it is known, proven on real defective parts across many logged runs… and lived in no document an agent could read.

## What

A third workflow skill, `b123d-repair`, wired identically to `b123d-modeling`/`b123d-drawing`:

- **`src/build123d_mcp/skills/b123d-repair/SKILL.md`** — the cookbook (details below)
- `install_skill(skill="repair")` + SKILLS entry (claude / agents-md / cursor / windsurf targets all work)
- **`build123d://skill/repair`** resource
- `_INSTRUCTIONS` mention (1181 bytes total — under the 2 KB cap the test enforces)
- README + llms.md pointers
- **The `export()` gate-FAIL messages now point at the skill** — discovery at the moment of need, when the agent is staring at a failure it doesn't know how to fix

## The cookbook

Structured as **diagnose → escalation ladder → verify**, keyed by defect class from the gate output:

- **Step 0 — diagnose before cutting**: read the FAIL reason (BRepCheck / open edges / non-manifold / mesh-only each route differently), `locate_gate_defects()` for coordinates, per-face `BRepCheck_Analyzer` sweep to localize, and the *inherited vs self-introduced* fork (self-made defects want `restore_snapshot()` + reconstruction, not surgery).
- **Step 1 — the ladder, cheapest first**: `ShapeFix_Shape` → clean-boolean re-computation → `BRepAlgoAPI_Defeaturing` → four-stage sliver-face surgery (ruled-face replace → drop+tolerant-sew → patch+small-sew → thin-box local cut) → the mesh-gate-only variant for unmeshable ~zero-area faces. **Every rung states when it applies, when it fails, and which rung to try next.**
- **The two observed silent-failure traps are mandatory checks**, not footnotes: defeaturing silently filling an internal bore (+14% volume — the failure that shelved #303's auto-heal), and booleans silently dropping the invalid solid.
- **Step 2 — verify with the export gate**, never `validate()` alone: covers both the in-memory fake-heal round-trip problem and the large-shape fast-fallback screen.
- **Step 3 — heal-FIRST for editing an invalid import**: get a gate-PASSing baseline snapshot *and* export on disk before attempting the edit, so a valid artifact always exists.
- **Step 4 — avoidance**: the constructions that create these defects in the first place (exactly-coincident faces, tangencies, disjoint pattern fields).

Every recipe traces to a successful heal on a real defective part; collectively they cover the in-session heals behind a 100%-validity 81-fixture run (including all three known BRepCheck-invalid inputs).

## Tests

- Repair variants of the install_skill suite: `_load_raw`, claude/cursor/agents-md installs, three-skill coexistence in shared files, `dest_exists` independence, resource registration, `_INSTRUCTIONS` needle — **36/36 `test_install_skill.py` pass**
- Full suite: **959 passed**; the 11 initially-failing tests reduce to 9 path-traversal tests that **fail identically on clean `main` in this environment** (checkout under `/private/tmp`, where macOS's `/tmp` symlink trips the traversal guards) — verified by stash/checkout-main/re-run before pushing; unrelated to this change
- ruff + mypy + `ruff format --check` clean

## Related

- #382 — follow-up design for a *mechanism*-level `recover()` (this PR is the judgment layer; sequencing per that issue is skill-first, observe a run, then decide)
- #303 — the shelved auto-heal whose post-mortem ("the honest shape of this tool is advisory") this implements the advisory half of
- #381 — validate()'s large-shape accuracy gap, the other half of the validity story

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Added
+
+- **`b123d-repair` skill: a validity-gate repair cookbook (`build123d://skill/repair`, `install_skill(skill="repair")`).** Until now the server could *diagnose* an invalid solid (`validate()` reasons, `locate_gate_defects()` coordinates) but offered no guidance on *fixing* one — and field runs show agents burning entire sessions failing to heal a BRepCheck-broken import that a known recipe repairs in minutes. The skill distills field-proven repairs from real defective parts into a defect-class-keyed escalation ladder: diagnose from the gate output first, then `ShapeFix_Shape` → clean-boolean re-computation → `BRepAlgoAPI_Defeaturing` → four-stage sliver-face surgery (ruled-face replace → drop+tolerant-sew → patch+small-sew → thin-box cut), plus the mesh-gate-only variant for unmeshable ~zero-area faces. Every rung states when it applies, when it fails, and what to try next, with the two observed silent-failure traps as mandatory checks (defeaturing filling an internal bore at +14% volume; booleans dropping the invalid solid). Includes the heal-FIRST workflow for editing an invalid import (a valid baseline exists on disk before the edit starts) and an avoidance section for the constructions that create these defects (exactly-coincident faces, tangencies). The `export()` gate-fail message now points at the skill — discovery at the moment of need.
+
 ## v0.3.68
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## v0.3.69
 
 ### Added
 
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ When using an AI to write build123d scripts, the AI writes blind — it cannot s
 - `repair_hints` — get targeted fix suggestions for a given `execute()` error message
 - `workflow_hints` — guidance on using the tools effectively
 - `script` — assemble a reproducible Python script from the session's executed code blocks
-- `install_skill` — copy a b123d workflow skill (modeling or drawing) into the current project
+- `install_skill` — copy a b123d workflow skill (modeling, drawing, or repair) into the current project
 
 ## Resources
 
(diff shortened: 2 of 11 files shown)


pzfreo:
Ran an 8-angle adversarial review against this branch and pushed a follow-up commit (47cdf99) addressing everything that survived verification — 8 findings, all fixed:

1. **`OCP.BRepFill` was missing from the sandbox allowlist** — rungs 4.1/4.3's ruled-face/patch recipes were unusable in a default (sandboxed) deployment, exactly where the harder rungs matter most. Added to `security.py`.
2. **Rung 4.2's drop+tolerant-sew snippet never produced a result** — imported `ShapeFix_Solid` but never called it, called `sew.Perform()` but never retrieved `SewedShape()`. Completed the shell-extraction + `SolidFromShell` chain, smoke-tested against a real shape.
3. **The repair-skill pointer was only on `export()`'s FAIL messages** — `validate()` FAILs are far more common (610 vs 214 calls in the last full sweep) and hit first per the documented workflow. Added the same pointer to `_validate_report`.
4. **Rung 1 (global `ShapeFix_Shape` as first resort) contradicted the standing "prefer targeted solid repair over broad shape healing" rule** that #379 added to default_prompt.md/llms.md/the modeling skill one day earlier. Added a caveat reconciling the two.
5. **CLI help text excluded the new skill** — `--skill` help and the `--help` epilog still said "drawing or modeling" even though `--skill repair` already worked. Derived both from the `SKILLS` registry.
6. **The Step 0 diagnosis snippet built a fresh `BRepCheck_Analyzer` per face**, contradicting `_locate_subprocess.py`'s own idiom (one analyzer — the whole reason that check runs out-of-process, "can run for minutes on a complex part"). Fixed to one analyzer over the whole solid.
7. **CHANGELOG's `## Unreleased` heading** doesn't match this repo's convention or its own CLAUDE.md release procedure. Changed to `## v0.3.69`.
8. **The no-personal-markers test hardcoded `("drawing", "modeling")`**, so the new skill silently escaped that invariant. Derived from the `SKILLS` registry.

Full suite: 961 passed (up from 959), same 9 pre-existing path-traversal failures as clean `main` in this environment, unrelated to this PR. ruff + mypy + format clean.

PR #384, opened 2026-07-06
test: warm build123d before design audit baseline

ahfoysal:
This adds a lightweight build123d warm-up step before design audit timing so timing checks are stable and less flaky.

It also keeps the same test coverage and command behavior and is scoped to the validation/export path.


pzfreo:
Closing as superseded — this is byte-for-byte identical to what already merged via #380 (both `_design_audit_subprocess.py`'s `_warm_build123d_import()` and the new test file match current main exactly). Looks like #380/#384/#385/#379 were all opened within minutes of each other on the same content; #380 and #379 merged first. Thanks for the contribution.

PR #385, opened 2026-07-06
docs: validate export gate gotchas

ahfoysal:
Adds docs guidance around export gate behavior and validation expectations for #320-era workflows.

No API changes; this is a docs-only hardening for contributor discoverability.


pzfreo:
Closing as superseded — this content landed via #379 (merged in v0.3.69) before this PR was updated, and it's now fully duplicated: every distinctive phrase in this diff already matches current main verbatim or in a lightly refined form (a couple of small wording tweaks from #386's follow-up pass). Nothing here to cherry-pick. Thanks for the contribution — apologies for the overlap, this looks like it and #379 were opened around the same time on the same content.

Issue #360, opened 2026-07-04
validate()/measure()/etc. (_GEOMETRY_TIMEOUT class) kill the worker on large solids — same bug as #357, easier fix than #359


pzfreo:
## validate() (and the whole _GEOMETRY_TIMEOUT class) kills the worker on large solids — same failure class as #357/#359, but easier to fix than execute()

Follow-up to #357/#358 (render_view, fixed) and #359 (execute(), harder — stateful). Fresh
field evidence from a real 81-fixture CADGenBench sweep (`gpt55-v0366-xhigh-full-vtkfix`,
mcp 0.3.66 — i.e. **after** the #358 render fix, so this is a distinct recurrence, not the
same bug reappearing) shows the identical session-destroying pattern on `validate()`.

## Evidence

- **Fixture 202** (editing, 4.38M mm³ imported solid): `validate()` on a healed/re-sewn
  replacement face → `Error executing tool validate: Operation 'validate' timed out after
  60s. The worker was killed and restarted — all session state ... has been lost.`
- **Fixture 248** (editing, 33M mm³, 1234-face solid): the **same** `validate()` timeout hit
  **three separate times** across the fixture's run, each destroying all in-progress work.

Both are large solids, both hit exactly 60s — `worker.py`'s `_GEOMETRY_TIMEOUT = 60`, which
`validate` shares with `measure`, `verify_spec`/`suggest_spec`, `clearance`,
`cross_sections`, and `save_snapshot` (`@_op(..., _GEOMETRY_TIMEOUT)` on all of them). So
this isn't validate-specific — it's the whole shared-budget class, and `validate()` is just
the instance that happened to fire in this sweep. Same root cause as #357/#359: an
uninterruptible native OCC call (here, `BRepCheck`/mesh validity on a large B-rep) that
exceeds its budget is stopped by killing the worker, per the same by-design policy
documented in `security.py`.

## Why this one should be EASIER to fix than #359 (execute())

`execute()` is stateful (persistent namespace mutation across calls), which is why I
flagged it as needing a harder architectural fix in #359. Every op in the
`_GEOMETRY_TIMEOUT` class is different: they're all **read-only computations on the
current/named shape** — take a shape, run one native analysis, return a JSON result. No
namespace mutation, no multi-call state to preserve. That's structurally identical to what
`render_view`'s fix already solved (`_tessellate_shapes_bounded`: serialize shape(s) to a
[...]

PR #386, opened 2026-07-06
fix: validate() catches open edges on large shapes; no more silent false PASS (#381)

### LINKED ISSUE #381: validate()'s fast-fallback mesh check is structurally blind to open edges — misleads the agent mid-session on large shapes

pzfreo:
Fixes #381.

## The bug
`validate()`'s exact topology-stitch mesh check, when it overran its in-loop budget on a large shape, fell back to the fast coordinate-weld — which **only counts non-manifold edges and pins `mesh_open_edges` to 0**. So any shape big enough to punt reported a **guaranteed, unwarned false PASS** for the open-edge (unclosed-boundary) class — the single most common defect the gate exists to catch.

Caught live by a CADGenBench sweep (`gpt55-v0368-xhigh-full`): fixture 240's 15 `validate()` PASSes all carried `mesh_check: "fast"`, while every `export()` failed with the same "4 mesh open edge(s)" the agent's own verify loop never saw.

## The fix — adopt `export()`'s architecture exactly (what #381 asks for)
`validate()` now keeps its cheap **B-rep checks in the worker** and isolates **only** the expensive mesh stitch, running it in the same hard-bounded subprocess `export()` uses (`_run_mesh_gate_subprocess`, the exact check, same 80k-triangle ceiling). Consequences:

- **Open edges are caught in-loop** for a large shape (the exact check runs, not the blind fast fallback).
- **A subprocess timeout can never lose the verdict** — because the verdict was never inside the killable child. It degrades to `mesh_check: "skipped"` + an explicit *"mesh not verified"* warning (test-export before trusting) while the in-worker B-rep verdict stands.
- **Faster** in the degraded large-shape case (B-rep runs once, in the worker): an 847-face stress part goes ~57s blind-and-wrong (naive fix) → **~18s honest `skipped`**.
- The remaining in-worker fast fallback (small dense parts, where a STEP round-trip would dominate) now carries the same open-edge-unverified warning.

`validate()` no longer routes through `run_bounded_shape_op` (which isolated the *whole* gate, so a hard-kill lost the B-rep verdict); it mirrors `export()` instead.

## Verification
- New `tests/test_validate_open_edge_381.py` (8 tests): the fast-fallback warning; **positive open-edge detection on a real open shell**; large shape isolates only the mesh (`exact-subprocess`); small shape stays in-worker; **graceful `skipped` degradation keeps the B-rep verdict**; unserialisable-shape skip; end-to-end through the real subprocess.
- Latency measured on a genuinely large (847-face) part before/after; degraded case is faster *and* honest.
- Full suite: **974 passed**.

## Review
Adversarially self-reviewed by three parallel agents (correctness, subprocess/resource safety, test coverage). Their findings drove this design: the correctness + safety reviewers both identified that running the whole gate in the killable subprocess (my first approach) could lose the B-rep verdict on a hard-kill — which is exactly what the export-style architecture (B-rep in worker, mesh isolated) fixes. Test-coverage reviewer's gaps (the `skipped` branch and real open-edge detection) are now covered.

## Note / open question for the maintainer
Above the 80k-triangle mesh ceiling a shape is reported `skipped` (mesh not verified) rather than run through the old blind `fast` non-manifold check. This **matches `export()`** (which also skips above 80k) — the alignment #381 wants — but it does mean a >80k-tri shape with a *mesh-non-manifold* defect is now surfaced only at `export()`, not in-loop. Deliberate, and consistent with export; flagging it in case you'd prefer to keep a (warned) fast non-manifold pass for that band.


pzfreo:
Follow-up pushed: raised the mesh-gate triangle ceiling based on real measurement.

**The 80k ceiling was stale.** Measured directly: a 90,840-triangle part (13.5% over 80k) completes the full uncapped exact check in **5.8s** — it was being skipped for "too slow" when it was actually cheap. The `~0.3ms/triangle` estimate the ceiling was sized against was ~5× pessimistic. A scaling curve (90k/172k/326k/529k triangles → 5.8/13.2/32.1/68.1s, worst case including the open-edge ladder) grounds a new ceiling in real cost against the real available budget.

**Fix:** the isolated path (subprocess, `deadline=inf`, backstopped by an external `subprocess.run` timeout) gets its own, much higher ceiling — `_EXACT_ISOLATED_MAX_TRIS = 300_000` — comfortably under the default ~100s subprocess budget even in the worst (laddered) case. The in-worker `_EXACT_EXPORT_MAX_TRIS` (80k, used only by STL-only export) is left untouched: that path has no external kill switch, bounded solely by a 35s in-process soft deadline around an un-interruptible native call, so raising it would reintroduce the exact risk #360 fixed.

New regression test (`test_isolated_ceiling_covers_a_shape_above_the_old_inline_cap`) proves it end-to-end: a real ~90,840-triangle part now runs `mesh_check: "exact-subprocess"` instead of being skipped. Full affected suites + targeted new tests all green.

PR #388, opened 2026-07-06
fix: flatten successful single-solid STEP exports

faisal-shah:
## Summary

This PR tightens STEP export around one rule:

_A STEP export that contains exactly one solid should open as a part, not as a one-component assembly._

The previous fix in #356 handled the failure path I had seen at the time. With build123d 0.11, `export_step()` could raise for imported or reimported single solids. The fix reconstructed the solid and retried the CAF writer, which kept the STEP output as a single product and avoided the `Compound` workaround that creates `NEXT_ASSEMBLY_USAGE_OCCURRENCE` records.

While regenerating downstream CAD packages, I found a second path in the same issue class. Some fresh one-solid wrappers, especially located primitives and located `Part` wrappers, do not raise. The high-level writer returns success, but the resulting STEP file still contains `NEXT_ASSEMBLY_USAGE_OCCURRENCE`. SolidWorks and similar CAD kernels then open the file as a one-component assembly rather than a part. Since the writer returned success, the previous fallback was never reached.

This PR changes the rule from "handle the known exception path" to "verify the written artifact before returning." `_write_step()` now counts the solids before writing. For single-solid exports, it only accepts high-level writer output if the STEP file has no assembly usage records.

If the high-level writer produces a one-component assembly for one solid, the fix does not immediately drop to the raw writer. It first bakes any non-identity location into the solid geometry, resets the TopoDS location to identity, and retries the CAF writer. That produces a flat single-product STEP while preserving names and colours in the normal fix path. The raw `STEPControl_Writer` remains only the last-resort fallback if CAF still cannot produce a flat single-solid file.

CAF here means Open Cascade's XCAF/OCAF document path, exposed through `STEPCAFControl_Writer`. In practical terms, that is the path that carries product names, labels, colours, and assembly structure into STEP. `STEPControl_Writer` is more basic and robust, but it writes geometry without that metadata. This PR keeps the metadata-preserving path for the cases where it can be made structurally correct.

## Why this should close the class

The underlying issue is not just that one writer call can fail. The broader issue is that a one-solid export can leave `export()` with assembly structure in the file.

The implementation now checks the final STEP structure before returning for every single-solid export path:

1. high-level writer succeeds and writes flat: return
2. high-level writer succeeds but writes assembly structure: bake the location into the single solid, retry the CAF writer, verify flat
3. high-level writer raises: reconstruct or bake the single solid, retry the CAF writer, verify flat
4. CAF retry still writes assembly structure or fails: fall back to raw `STEPControl_Writer`
5. genuine multi-solid export: keep assembly structure

That makes the behavior independent of which build123d wrapper type triggered it and independent of whether the high-level writer reports success or raises. It also keeps names and colours on the normal repaired path, so the fix improves part structure without unnecessarily regressing STEP metadata.

## Compatibility and tradeoffs

The normal repaired path should not make metadata worse. If CAF writes a bad one-solid STEP, the fix bakes the non-identity location into the solid, resets the TopoDS location to identity, and retries the CAF writer. That keeps the same world-space geometry, produces a flat part file, and preserves names and colours.

The main representation change is that the single solid no longer carries a separate location transform in the STEP structure. The transform is baked into the geometry instead. That is intentional: the separate location is what causes CAF to encode the solid as a component in a one-component assembly.

The raw `STEPControl_Writer` fallback can still lose CAF names and colours, but it is only used after the CAF path cannot produce a flat single-solid file. That was already the last-resort behavior for hard writer failures. This PR makes that path less likely, because located single solids now get a metadata-preserving CAF retry before falling back to raw geometry.

The priority order is unchanged: preserve geometry first, write the correct part-vs-assembly structure second, preserve names and colours when CAF can do so, and use the raw writer only when CAF cannot produce a correct flat part.

## Tests

Added regression coverage for the full behavior:

- import-derived single solids still export flat and preserve labels and colour metadata
- located primitive wrappers export flat and preserve labels and colour metadata
- located `Part` wrappers export flat and preserve labels and colour metadata
- a monkeypatched high-level writer that reports success while writing a one-component assembly is detected and rewritten flat
- real multi-solid compounds still export with assembly structure and preserve all solids

Commands run:

```powershell
uv run --native-tls --allow-insecure-host pypi.org --allow-insecure-host files.pythonhosted.org pytest tests\test_export_step.py -q
uv run --native-tls --allow-insecure-host pypi.org --allow-insecure-host files.pythonhosted.org ruff check src\build123d_mcp\tools\export.py tests\test_export_step.py
uv run --native-tls --allow-insecure-host pypi.org --allow-insecure-host files.pythonhosted.org ruff format --check src\build123d_mcp\tools\export.py tests\test_export_step.py
```

Full `pytest -q` also completed on my Windows host with `955 passed, 26 skipped`. The only failure was `tests\test_tools.py::test_export_symlink_escape_rejected`, which fails before reaching this change because the current Windows session does not have symlink creation privilege (`WinError 1314`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- **`export()` now rewrites one-component assembly STEP output even when build123d's high-level writer reports success.** The v0.3.65 fix covered the import-derived solid path where `export_step()` raises, but fresh one-solid wrappers such as located primitives can return success while still writing `NEXT_ASSEMBLY_USAGE_OCCURRENCE`. `export()` now checks the written STEP structure for every single-solid export, bakes any non-identity location into the one solid, and retries the CAF writer so the output stays flat while preserving names and colours. Genuine multi-solid compounds still export as assemblies.
+
 ## v0.3.69
 
 ### Added
--- a/src/build123d_mcp/tools/export.py
+++ b/src/build123d_mcp/tools/export.py
@@ -97,6 +97,48 @@ def _write_svg(shape, abs_path: str) -> None:
     exporter.write(abs_path)
 
 
+def _n_solids(shape) -> int:
+    try:
+        return len(shape.solids())
+    except Exception:  # noqa: BLE001 - no solid topology to reason about
+        return 0
+
+
+def _raw_step_write(shape, abs_path: str, *, single_solid: bool = False) -> bool:
+    from OCP.IFSelect import IFSelect_ReturnStatus
+    from OCP.STEPControl import STEPControl_AsIs, STEPControl_Writer
+
+    transfer_shape = shape.solids()[0] if single_solid else shape
+    writer = STEPControl_Writer()
+    writer.Transfer(transfer_shape.wrapped, STEPControl_AsIs)
+    return writer.Write(abs_path) == IFSelect_ReturnStatus.IFSelect_RetDone
+
+
+def _single_solid_step_is_flat(abs_path: str, n_solids: int) -> bool:
+    return n_solids != 1 or _nauo_count(abs_path) == 0
+
+
+def _flat_single_solid_copy(shape):
+    """Return one located solid as identity-location geometry for flat CAF STEP."""
+    from build123d import Solid
+    from OCP.BRepBuilderAPI import BRepBuilderAPI_Transform
+    from OCP.TopLoc import TopLoc_Location
+
+    solid = shape.solids()[0]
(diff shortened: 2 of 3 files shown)


pzfreo:
Merged, thank you! Nicely scoped fix and thorough test coverage — especially the monkeypatched writer test simulating the silent-success-with-assembly case, which makes the regression check independent of which build123d version actually exhibits the bug. Only had to resolve a trivial one-file CHANGELOG conflict against an unrelated PR that landed the same day; folded your entry into the existing v0.3.70 section.

PR #389, opened 2026-07-07
fix: correct regressions in b123d-repair skill from #387

### LINKED ISSUE #1: Security: remaining hardening items (code execution sandbox, resource limits, logging)
## Background

A security review of the MCP server identified the following issues. The path traversal and temp-file race condition have been fixed in a separate PR. This issue tracks the remaining items.

---

## High: No resource limits / denial-of-service

The `execute` tool places no bounds on memory, CPU, or execution time. A single call can exhaust the host:

```python
code = "x = [0] * (10**9)"  # OOM
code = "while True: pass"    # CPU spin
```

**Suggested fix:** Use Python's `resource` module to cap memory and add a `signal.alarm` timeout around `exec()`.

---

## High: Persistent namespace leakage between calls

The session reuses a single namespace across all `execute` calls. Variables (and any sensitive values) from one call are visible to the next.

**Suggested fix:** Document clearly that sensitive data must not be passed through `execute`. Consider whether a per-call namespace (with explicit state passing via `current_shape`) is feasible.

---

## Medium: No audit logging

No record is kept of which tools were called, with what arguments, or what errors occurred. Security incidents are undetectable.

**Suggested fix:** Add structured logging (tool name, timestamp, truncated params, result status) to a local log file.

---

## Medium: Error messages leak implementation details

Exceptions from underlying libraries are returned verbatim, potentially exposing internal file paths and module names.

```python
return f"Error: {type(e).__name__}: {e}"  # session.py:19
```

**Suggested fix:** Log the full exception internally; return a sanitised message to the caller.

---

## Low: No authentication

The server accepts any caller over stdio with no token or identity check. This is acceptable for local use but should be documented explicitly, and reconsidered if remote transport is ever added.

---

## References

- `session.py:19` — error message leakage
- `session.py:6-10` — shared namespace
- `tools/export.py` — (fixed) path traversal
- `tools/render.py` — (fixed) temp-file race condition

### LINKED ISSUE #3: UX improvements from real-world usage: assembly preview, rendering quality, richer measurements
## Background

Used the MCP server to model a purfling roller device (two-roller compression jig for lutherie). Overall useful for fast iteration, but hit several friction points. Raising these as a bundle since they're related.

### 1. Multi-object sessions (highest priority)

Currently only one shape can be active for rendering at a time (`result = shape`). When modelling an assembly, you lose the ability to see parts in context — e.g. checking whether a roller fits correctly inside a U-frame.

**Suggestion:** a `show(name, shape)` registry so multiple named objects can be accumulated in the session, with `render_view` able to show all of them (or a named subset) composited together.

---

### 2. Rendering quality

The triangulation tessellation produces visible artefacts (a "sunray" radial pattern on flat circular faces, streaking on cylindrical surfaces). This makes it hard to inspect small features like bores and slot edges.

**Suggestion:** increase tessellation quality, or offer a quality parameter on `render_view`. A cross-section / slice plane view would also be very useful for checking internal geometry (bores, wall thicknesses) without needing to export.

---

### 3. Richer measurement queries

`bounding_box` is useful but limited. During this session I wanted to verify:
- axle slot width and depth
- wall thickness on either side of a slot
- clearance between two named bodies

**Suggestion:** extend `measure` to support at minimum `min_wall_thickness` and face/edge distance queries. Even a simple `measure("volume")` would help catch missing booleans.

---

### 4. Session snapshots

A reset wipes all variables. When iterating on a complex assembly it would be useful to checkpoint a known-good state and restore it if an experiment breaks things.

**Suggestion:** `save_snapshot(name)` / `restore_snapshot(name)` calls.

---

### 5. Multi-format export

Currently one format per `export` call. STEP preserves exact geometry for downstream CAD tools; STL is needed for slicers and GitHub preview.

**Suggestion:** accept a list of formats, or add a convenience `export_all(basename)` that writes both `.step` and `.stl` in one call.

---

## Priority order

1. Multi-object / assembly preview — catches fit and clearance problems early
2. Rendering quality — directly affects ability to inspect geometry
3. Richer measurements — reduces need to export just to check a dimension
4. Snapshots — quality of life for complex sessions
5. Multi-format export — minor convenience

🤖 Raised via [Claude Code](https://claude.com/claude-code)

pzfreo:
## Why

#387 (merged) refined the `b123d-repair` skill using field evidence from three hard fixtures, but wasn't tested broadly enough before merging. An xhigh-effort workflow-backed adversarial review of the merged PR — 6 independent finder angles plus independent verification of every candidate against real OCP bindings, not just plausible-sounding reasoning — found **9 confirmed defects**. This PR fixes all of them.

## What was wrong

1. **`Solid(TopoDS.Solid_s(...))` crashed with `Standard_TypeMismatch`** whenever the raw OCCT result was a `Compound` rather than a bare `Solid` — the normal case for anything built with build123d's own operators (even a plain `Box()`), not just imported STEP files. Rungs 3-4 hit this on every non-single-solid-import use — a regression worse than the bug it replaced (the old `Part(...)` at least ran, it just silently under-reported volume).
2. **Rung 1 and Rung 2 were never actually switched off `Part(...)`**, even though the CHANGELOG claimed "every rung now uses the robust wrapper."
3. **Rung 4 option 1 silently dropped inner wires** — a face with a hole (outer wire + inner wire) got rebuilt as a solid plate via `TopExp_Explorer(...).Current()`, `IsDone()` True, no error, hole gone.
4. **Step 2's new "stealth non-fix" warning contradicted rung 4 option 1's own design** — that option deliberately leaves volume/topology/bbox unchanged (same wire, same boundary), which is exactly the signal Step 2 told the agent to distrust.
5. **Rung 4 option 1's `healed` was left unbound** on the documented failure path (non-planar wire), so a later reference raised `NameError` instead of the clean "move to option 2" signal the prose promised.
6. **Rung 4 option 2 (`BRepFill_Filling`) had no `IsDone()` check at all**, unlike its sibling option 1, despite the very next sentence in the same PR noting this call can silently produce a bad face.
7. **The old ruled-face applicability gate was deleted without being restated** for its replacement (`BRepFill_Filling`) — an agent could waste an attempt on a face whose wire is itself broken, not just its surface.
8. **The promised "small-tolerance re-sew" had no matching code anywhere** — the only sewing snippet in the rung used a much larger, differently-purposed tolerance.
9. Cleanup: manual `TopExp_Explorer` edge-collection reinvented `Face.edges()`.

## What changed

- One shared `as_solid()` helper, defined once in Step 1 and reused by every rung, that dispatches on the raw shape's actual type (`Solid` / `Compound` / `Shell`) and always returns a build123d `Solid` with a correct volume. This fixes #1 and #2 at the same time — no more per-rung wrapper drift.
- Rung 4 option 1 now rebuilds from `outer_wire()` plus explicit inner `wires()`, raises explicitly on `IsDone()` False instead of leaving `healed` unbound (fixes #3, #5).
- Rung 4 option 2 gets the same explicit `IsDone()` guard, its applicability gate restated (fixes #6, #7), and the actual small-tolerance re-sew snippet (fixes #8).
- Step 2's warning now carves out an explicit exception for rung 4 option 1 (fixes #4).
- Switched to `Face.edges()`/`.outer_wire()`/`.wires()` throughout instead of manual `TopExp_Explorer` (fixes #9).

## Verification

This time, every changed **and unchanged** code block in the file was extracted verbatim with a regex and `exec`'d against real build123d/OCP shapes — not re-typed by hand, which is exactly the gap that let some of the original findings through undetected:
- Every rung tested against **both** a bare imported `Solid` and a `Compound`-wrapped shape (e.g. `Box() - Cylinder()`).
- Rung 4 option 1 tested against a genuine multi-wire face (a plate with a drilled hole) — confirmed the old code silently fills the hole (volume 1858.6 → 1883.0), confirmed the fix preserves it exactly (1858.6 → 1858.6).
- Rung 4 option 2's fill + re-sew sequence tested end-to-end — confirmed the fill alone leaves a real gap (666.67 vs 1000 on a test box) and the re-sew snippet actually closes it (→ 999.999...).
- 36/36 `test_install_skill.py` pass; full `pytest -q` green; `ruff check`/`ruff format --check` clean.

## Related

Follow-up to #387. No CHANGELOG version bump — v0.3.70 is still unreleased (`0.3.70.dev0`), so the existing entry was amended in place to describe the corrected work rather than adding a confusing "fix of a fix" note.


pzfreo:
Pushed a follow-up commit after running a second xhigh adversarial review against this branch — it found 4 more confirmed defects in the first round of fixes, most seriously a **segfault** (calling `.Add()` on an already-failed `BRepBuilderAPI_MakeFace` in rung 4 option 1 exits 139, not a catchable exception) and a silent-partial-volume bug in `as_solid()` itself when the raw result is a multi-solid `Compound` (reproduced: fusing two disjoint boxes silently kept only one, ~1000 instead of ~2000).

All fixed, and this time every code block — changed and unchanged — was extracted verbatim from the file with regex and `exec`'d against real shapes, including new cases (a genuinely multi-solid Compound, a non-planar face that must raise without crashing) that weren't tested the first time around. Full test suite green. Still unmerged, still awaiting review.

pzfreo:
Pushed one more fix: `as_solid()` checked solid-count and shell-count independently, so it still silently dropped topology when a compound held one solid **plus** a stray shell/face — returning as soon as it found exactly one solid without checking whether the compound also held other, unexamined content. Reproduced: `Compound([Box(10,10,10), a sewn 5x5x5 shell])` has volume ~1125, but the helper returned only 1000.

Rewrote it to walk direct children only (`TopoDS_Iterator`, not the recursive `TopExp_Explorer` used before), recursing into nested compounds, and raise on **any** mix — not just >1 of the same type. Re-verified the entire existing regression suite still passes (every rung, both wrapper types, the multi-wire face, the two must-raise-not-crash/not-silently-drop cases from the last round) plus this new case. Full test suite running now, will confirm.

PR #393, opened 2026-07-07
fix: b123d-modeling skill gains a boss-extension recipe

pzfreo:
## Why

The `b123d-modeling` skill already warns that additive edits with exactly-coincident faces often fail to fuse, and suggests "interpenetrate slightly, bury the added feature into the base." Extending an existing boss or cylindrical feature — a constant cross-section, so any add-on has to fuse against an exactly-matching boundary — is the single most common real instance of that trap, and interpenetrating doesn't reliably fix it.

## Field evidence

A CADGenBench editing fixture asked to extend a hollow boss to reach a target height. The agent tried **five different bridging/overlap fuse variants** against a separately-built cylinder — direct overlap, no-overlap trial, a rounded-top replacement, a center bridge, side-wall bridges, near-wall bridges — every one failing with the same non-manifold or two-disjoint-solids signature. It eventually gave up on that boss entirely and found an easier target elsewhere in the part, rather than ever actually fixing the extension.

Two independent past sessions had already found and verified the actual fix for this — it just never made it into the shipped skill.

## What changed

Added both variants of the same underlying idea — extrude the boss's *own* end-cap face instead of adding a separately-built cylinder or tube. A face extruded from the boss's own boundary shares the exact underlying geometry with what it's fused against, so the boolean dissolves the shared boundary cleanly; a coincidentally-matching new primitive does not, even positioned perfectly.

- **One-sided** (extend toward a target on one end only): `BRepFeat_MakePrism` on the seating face — a prism *feature* that extends the existing topology in place, so there's no second solid to fuse at all.
- **Symmetric** (both ends grow by the same amount): plain build123d `extrude()` of each end-cap face by half the growth, then fuse — no raw OCCT needed.

## Verification

Both snippets were extracted verbatim from the file (not retyped) and `exec`'d against a synthetic hollow boss (annular cross-section, coaxial bore, built through a real STEP export/re-import round-trip so it matches an imported-solid target, not a native build123d primitive):

- One-sided: exact volume match, exact new top-Z, `BRepCheck_Analyzer` valid.
- Symmetric: exact volume match, exact new top-Z, `BRepCheck_Analyzer` valid, face count unchanged (confirming the coincident boundary was genuinely dissolved, not left as a duplicate).

I also tried to reproduce the *naive* fuse-a-new-cylinder failure on the same idealized synthetic case — it actually succeeds there. That's consistent with the rest of this skill's evidence bar ("verified on a real defective part," not a trivial synthetic reproduction) — the failure needs the imperfection/complexity of an actual imported casting, which is exactly the environment the field evidence came from.

36/36 `test_install_skill.py` pass. Full `pytest -q` running; will confirm.


pzfreo:
Broadened this while it was still open — a second, independent fixture (raise an annular bore opening by 10mm) hit the exact same trap live: three straight construction attempts (overlap fuse, no-overlap fuse, patch-the-exported-defect) each passed `validate()` in memory but failed `export()`'s stricter round-trip check, before switching technique to extrude-the-actual-face and succeeding immediately. Same fix, same code, just widened the prose from "boss/cylindrical feature" to any relocated planar/annular face — that's what `BRepFeat_MakePrism` on an existing face actually covers. Both code blocks re-verified unchanged.

pzfreo:
The xhigh adversarial review caught 3 real bugs — thanks. All from the same root cause: the first draft only tested a single Z-aligned, single-end scenario.

1. **One-sided snippet hardcoded `gp_Dir(0, 0, 1)`** instead of deriving direction from the seat face's own normal. Verified: on an X-aligned boss, this silently collapses the result to zero volume with `IsDone()` still `True` and `BRepCheck_Analyzer` still reporting valid — no error signal at all. Fixed by using `seat.normal_at(seat.center())`.
2. **Symmetric snippet's "repeat for the other end" fused onto the original `part`** instead of the already-extended result, silently discarding the first end's growth. Rewrote as one complete worked example (both ends, chained `.fuse()` calls) instead of a "repeat this line" instruction.
3. **Same snippet extruded by the full `delta` instead of half**, despite the prose saying otherwise — doubling the final length. Fixed by explicitly computing `half = delta / 2`.

Also added the `IsDone()` guard the review flagged as inconsistent with `b123d-repair`'s established convention.

Re-verified end-to-end: Z-aligned boss (regression), X-aligned boss (bug 1), and a real both-ends extension (bugs 2-3) — exact volume and growth match in every case. Full test suite running, will confirm.

PR #397, opened 2026-07-07
fix: exact mesh gate no longer silently welds off-vertex tessellation nodes

pzfreo:
## Why

Cross-validated our own gate's PASS verdict against CADGenBench's own official mesh sanity checker (run directly from its source — `cadgenbench.common.mesh.tessellate_and_validate` — not the packager's known-broken `--official-sanity-check` path) and found a real submission our gate passed cleanly (`0 mesh non-manifold edges, exact-subprocess, 0 open edges`) that the official checker rejected:

```
vertex #492: an edge endpoint node is 1.993e-01 mm from the vertex
(> deflection 0.005717647362146476 mm); edge polygon orientation is
inconsistent with FirstVertex/LastVertex
```

## Root cause

The gate's vertex-merge step (closing fillet/cone/pole apices where an edge's tessellated endpoint should coincide with its BREP vertex) unioned every candidate node into the vertex **unconditionally**, with no check that the node was actually anywhere near it. A patched/healed face (a sliver sew, a tolerance-fudged repair patch) whose polygon endpoint missed its vertex by a fraction of a millimetre got silently welded into "the same point" instead of flagged. I traced it to the exact location in the field case: two ~1.5mm² and ~5mm² BSpline sliver faces at a healed junction — the geometric signature of a repair patch, not the original import or the edit itself.

## What changed

Added the same guard CADGenBench's own checker uses (confirmed by reading its source directly, not guessed): before unioning a candidate node into a BREP vertex, verify it is within the mesh deflection of the vertex's analytic position; report each vertex that fails as a new `mesh_vertex_deflection_defects` count rather than merging it silently.

- `_mesh_defects_exact()`'s return tuple gains a 6th element, threaded through all 11 return points.
- Wired into `_gate_report`'s pass/fail decision and `reasons`.
- Threaded through `_gate_subprocess.py` / `_run_mesh_gate_subprocess` so `export()` on a large part (out-of-process mesh check) sees it too.
- Added to `locate_gate_defects()` as a new `mesh_vertex_deflection_defect` kind with the vertex's coordinates and measured deviation (needed its own explicit `BRepMesh_IncrementalMesh` call — a freshly-imported shape has no existing triangulation to read).
- `package_submission.py`'s (cadgenbench-build123d) `proxy_gate()` default fallback tuple will need the matching update once this version is adopted there — noted, not part of this repo.

## Verification

Ran against the real submission that exposed the bug: the fix reproduces CADGenBench's own verdict exactly — 2 defective vertices located, one of them the reported vertex #492, `passes_gate` now correctly `False`.

No regressions: full existing exact-mesh-check test suite passes (clean solids, fillets, a sphere's poles/seams, disjoint touching bodies — all still 0 false positives). `test_mesh_defects_exact_no_false_nm_vertex_on_clean_solids` extended to also assert `vdefl == 0` on those same clean shapes.

**On a synthetic true-positive test**: I attempted a from-scratch minimal reproduction but it proved unreliable — `BRepMesh_IncrementalMesh` silently rebuilds a face's triangulation from scratch on its next call even at an identical deflection once the shape has been touched by `bounding_box()` in between, discarding any hand-edited node before the check under test ever sees it. Given the field-case verification is cross-checked against CADGenBench's own independent implementation, it's stronger evidence than a contrived synthetic case would be anyway — happy to take another pass at a synthetic repro if reviewers want one.


pzfreo:
Ran a detailed adversarial review on this branch (8 independent finder angles + cross-checks against live CI logs and CADGenBench's actual source). Found one CI-blocking bug and 7 real completeness gaps, all fixed in the latest commit:

1. **[Blocking]** Stale return-type annotation broke `mypy` in CI (15 errors) — `_mesh_defects_exact` still said 5-tuple after every return site became a 6-tuple.
2. The deflection guard was only applied to the base non-manifold pass — the open-edge ladder's own independent vertex-merge (`_open_pass`) still unioned unconditionally, so a defect invisible at the base deflection but exposed only at a finer escalated rung could still slip through. Same guard now runs there too, with a test that isolates this specific path.
3. The new locator function had no size/time budget, unlike its gate-side counterpart — added a triangle-count ceiling.
4. Tried making the locator reuse the other locators' shared tessellation pass — measured directly that this balloons one real part's triangle count 14x (the gate's coarser angular tolerance is deliberate, not an oversight). Reverted; kept the triangle budget as the actual safety net instead.
5. The locator used Euclidean distance where the gate (and CADGenBench's real algorithm) uses max-per-axis — fixed, since the mismatch meant the locator could flag a vertex the authoritative gate doesn't actually fail on.
6. The locator reintroduced a slow edge-adjacency pattern `_edge_face_adjacency` was specifically written to avoid — switched to reuse that helper.
7. No positive-case test existed anywhere for this defect class. Real malformed geometry that reproduces it resists synthetic construction (sewing heals mismatches geometrically; hand-edited triangulation nodes get discarded on the next mesh pass) — added tests that instead drive the exact `BRep_Tool.Pnt_s` lookup via a targeted monkeypatch, deterministic and confirmed against the real field case's numbers.
8. The `b123d-repair` skill's Step 0 diagnose table had no entry for the new FAIL reason — added it plus a matching Rung 5 recipe.

No regressions: 978 passed (975 baseline + 3 new tests), same 9 pre-existing environmental failures as every run this session, `mypy` clean.

PR #409, opened 2026-07-09
fix: automatic in-process fallback when the worker subprocess can't start

faisal-shah:
#143 added `--in-process` as a manual fallback for hosts that can't run the worker subprocess. This makes it automatic, and cuts the time to detect the failure from 60s to ~20s.

`WorkerSession` degrades to in-process execution when `_start_worker()` fails, at construction and on restart, instead of raising and leaving every worker-backed tool returning `Worker was not running; restarted` for the life of the process. It builds the `Session` in-process (same path as `InProcessSession`), replays the parent-side `execute()` history so state survives, warns once on stderr, and keeps serving. Opt out with `BUILD123D_NO_WORKER_FALLBACK=1`.

## Why it's worth merging

The failure is unrecoverable on the affected host, not a config nudge the operator can act on. On the Copilot/Codex CLI on Windows the worker spawns but never finishes multiprocessing bootstrap: it sits idle at ~10 MB, never imports build123d, and is killed at the ready timeout. I confirmed the identical code spawns and runs in ~0.3s outside that host, including from a threadpool thread under a live asyncio loop, so the fault is the host's process sandbox breaking the spawn handle handshake, not our spawn usage or threading model. Where the worker cannot run, hard-failing every call helps no one; `--in-process` is already the answer, and this stops requiring the operator to diagnose a cryptic per-call error and set it in advance. On hosts where the worker starts, behavior is unchanged: the path is dead unless `_start_worker()` fails.

## Two-phase startup (the 60s to ~20s part)

Detection cost mattered because the failing restart waited out the entire ready timeout before falling back. Startup is now split. The worker sends `{"alive": True}` the instant it enters `worker_main`, before it imports or builds anything; the parent waits a short `_WORKER_ALIVE_TIMEOUT` for that ping, then the longer `_WORKER_READY_TIMEOUT` for the Session build. The #143 hang is a spawn-bootstrap stall that never runs `worker_main`, so it never pings and is caught in the short window.

The 60s was a conservative default from the original persistent-worker commit and was never tuned. It never needed to cover build123d/OCP, which load lazily on the first `execute`, not at worker startup. The alive path is pure-Python imports (`import build123d_mcp.cli`, measured ~0.2s), so a 20s budget has wide margin, and a slow-but-healthy cold import or `--library` index still gets the full ready budget in phase two.

## Behavior

- The degrade is sticky for the session: once the worker proves unrunnable, the session stays in-process. Spawn breakage is a host property, not a transient.
- State is rebuilt by replaying the parent-side `execute()` history into the in-process `Session`, bounded by the same wall-clock budget as the worker-restart replay and keeping the rebuilt prefix. Snapshots and tool-imported geometry are not restored, matching the existing restart contract.
- The in-process dispatch body is shared between the auto-fallback and explicit `InProcessSession`, so the two paths cannot diverge.

## Tests

- `tests/test_worker_inprocess_fallback.py`: startup degrade, restart-path degrade with history replay, replay budget bound, error-contract parity with the worker path, `reset` after degrade, opt-out env var.
- The real-worker suites (`test_worker_replay`, `test_worker_session_tools`, `test_worker_boundary_coverage`, `test_worker_concurrency`) exercise the new two-phase startup end to end; `test_worker_timeouts` stubs it, and its stub is updated for the new attributes.
- Green across all of the above plus `test_in_process_session`. `ruff` clean. `mypy` unchanged (only the pre-existing Windows `resource` attribute warnings).


faisal-shah:
Superseded. Narrowing this to the one change that stands on its own — silencing build123d's log propagation to stderr (#143 noise on Windows). Reopened as a focused PR.

PR #411, opened 2026-07-09
fix: stop build123d INFO logs leaking to server stderr

faisal-shah:
build123d logs at INFO on every object construction (`"<ctx> context requested by <op>"`). build123d itself attaches only a `NullHandler`, but FastMCP installs a root stderr handler, so those records propagate to the server's stderr.

On a host whose stderr is a pipe where writes fail, that turns into noise in the user's output. Concretely, under the Copilot/Codex CLI on Windows each stderr write raises `OSError(EINVAL)`, and the standard-library logging machinery responds by printing a `--- Logging error ---` traceback — which lands in the middle of the tool result the user reads. A one-line `execute` that builds a `Box` comes back with a logging stack trace wrapped around it.

## Fix

`_build_session` — the setup path shared by the worker and in-process modes — sets `logging.getLogger("build123d").propagate = False`. build123d's per-object records stop reaching the root stderr handler. build123d keeps its own `NullHandler`, so anything that actually wants the log is unaffected, and setting the flag by logger name works before build123d is imported (it persists to when it first logs).

Placing it in `_build_session` covers both modes from one point, matching that function's existing "a setup step can't be added to one mode and forgotten in the other" contract.

## Scope

This is the stderr-noise half of the #143 experience on Windows, and it stands on its own: it also cleans up the output under `--in-process`, which is the practical way to run on a host where the worker subprocess can't start. It does not attempt to make the worker subprocess itself run under such a host — that's a separate problem.

## Tests

- `tests/test_build123d_log_propagation.py`: `_build_session` sets `propagate = False`, and a build123d INFO record no longer reaches an installed root stderr handler afterward.
- Green with `test_in_process_session`. `ruff` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.3.76
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


faisal-shah:
Abandoning. In the --in-process config this only silences benign stderr log-file noise, not anything user-visible on the happy path. Not worth the change.

PR #420, opened 2026-07-11
Add compact inspect_part checkpoint

### LINKED ISSUE #417: Add a generic feature-inventory audit helper for generation checkpoints
## Context
The best generation runs improved when the agent used recognizers and measurements as a structured audit rather than relying on visual impressions alone.

Examples:
- `logs/gpt55-v0377-xhigh-gen3-dominant/126.log`: recognizers found top holes that stopped blind and lateral spotfaces cut only about 1 mm; fixing this improved the result.
- `logs/gpt55-v0376-xhigh-full/126.log`: a final audit found only 25 of 33 intended split-flange holes were recognized; moving the hole pattern corrected the inventory.
- `logs/gpt55-v0377-xhigh-gen3-dominant/111.log`: stable checks for bbox, center bore, PCD, hole count, bosses, and cross sections protected the model while adding under-flange relief.

## Problem
Agents can call `measure()`, `find_holes()`, `find_hole_patterns()`, `find_bosses()`, and `cross_sections()` individually, but they have to invent the audit structure each time. That leads to inconsistent final checks and missed feature-count/depth defects.

## Proposal
Add a generic helper or documented pattern that produces a compact generation checkpoint audit:
- bbox/envelope
- solid count / topology summary
- hole groups by axis, diameter, depth, and pattern
- boss groups by axis, diameter, height
- cross-section area profile along a selected axis
- warnings for likely shallow/partial cuts, missing expected counts, or flat constant-section bodies

This could be a new MCP verb or a modeling-skill recipe if a full tool is too heavy.

## Acceptance criteria
- Agents can get the core feature inventory in one call or one documented code pattern.
- Output is compact enough to use inside generation loops.
- Tests cover holes, bosses, patterns, and a thin-walled/cored profile example.

## Non-goal
Do not encode benchmark fixture expectations. Expected counts should come from the agent's dimension table or be absent; the helper should be generally useful.

pzfreo:
## Summary

- add a read-only `inspect_part` checkpoint for bbox/topology, grouped holes and bosses, hole patterns, and cross-section variation
- support optional agent-supplied expectations with explicit `PASS`/`FAIL`, strict one-to-one feature-group matching, and actionable mismatch/warning output
- preserve bolt-circle and linear-pattern geometry plus member counts and diameters through shared pattern serialization
- isolate inspection of large shapes through the existing bounded STEP subprocess path
- reject unknown, overlapping, or malformed expectation schemas instead of silently weakening a verdict
- document the broader tool-surface consolidation separately for #421 without coupling that migration to this PR

## Review fixes

- renamed `feature_audit` to `inspect_part` before public release
- fail when one underspecified expectation absorbs multiple distinct actual groups
- reject duplicate/subset-overlapping expectation lines during validation
- normalize malformed JSON to a clear `ValueError`
- prove in-process/STEP-subprocess equality on through, blind, counterbored, bolt-circle, and boss features
- classify `inspect_part` in worker boundary coverage and register it in the MCP tool inventory

## Validation

- Ruff clean
- mypy clean across 61 source files
- full local suite: 1,039 passed, 1 skipped, with one stale tool-inventory assertion subsequently fixed
- post-fix affected suite: 40 passed
- feature-rich real bounded subprocess round trip covered

Closes #417

### DIFF
--- a/docs/tool-surface-consolidation-proposal.md
+++ b/docs/tool-surface-consolidation-proposal.md
@@ -40,9 +40,14 @@ one default product surface.
 
 ### Inspection
 
-- `inspect_part(object_name="", expected="", detail="")`
+- `inspect_part(object_name="", expected="", section_axis="Z", section_slices=7)`
 - `compare(a="", b="", kind="shape")`
 - `validate(object_name="")`
+
+The standalone `measure` MCP tool remains visible during the migration. The normal
+quick loop should use the existing in-`execute()` `measure(shape)` Python helper after
+each boolean; `inspect_part` is the heavier checkpoint, not a replacement for every
+fast measurement.
 
 ### Session
 
@@ -53,9 +58,22 @@ one default product surface.
 - `reset()`
 - `last_error()`
 
-This is approximately 12 to 14 tools rather than 38. The exact list should be
-confirmed by usage analysis before removal, but the categories and consolidation
-boundary should remain stable.
+### Drawing
+
+The drawing tools remain in the default surface until #421 makes an explicit,
+benchmark-backed decision about them:
+
+- `inspect_drawing`;
+- `view_axes`;
+- `lint_drawing`;
+- `render_drawing`;
+- `save_drawing_annotations`;
+- `suggest_view_layout`.
+
+Drawing is a real supported workflow with its own skill. It must not be silently
+dropped to meet an arbitrary tool-count target. Including it means the first
+consolidated surface will likely contain approximately 18 to 20 tools rather than the
+original 12-to-14 estimate.
 
 `validate` remains separate because "is this a valid solid?" is a frequent operation
 with a small, unambiguous contract. `compare` remains separate because before/after
@@ -67,7 +85,8 @@ comparison is central to editing and naturally operates on two objects.
 inspect_part(
     object_name="",
     expected="",
-    detail=""
+    section_axis="Z",
+    section_slices=7
 )
 ```
 
@@ -82,21 +101,12 @@ The default response is compact and always includes:
 - warnings;
 - an optional expectation verdict.
 
-`detail` is only for follow-up diagnosis:
-
-```text
-detail=""
-detail="features"
-detail="sections"
-detail="topology"
-detail="edit_candidates"
-```
-
-It expands evidence but does not alter verdict semantics.
-
-Public `section_axis` and `section_slices` parameters should be removed. The
-implementation should inspect principal axes automatically and use a stable, bounded
-sampling strategy.
+Do not add `detail` modes in the initial consolidation. Compact default plus a
+structured expectation language is already substantial; adding multiple diagnostic
+modes risks turning `inspect_part` into the god tool this proposal explicitly rejects.
+Follow-up evidence should first use the in-`execute()` analysis helpers. Automatic
+principal-axis selection may later remove `section_axis` and `section_slices`, but only
+after its reliability is demonstrated.
 
 ## Expectation contract
 
@@ -143,7 +153,6 @@ the verdict.
 
 The following operations become internal analysers used by `inspect_part`:
 
-- `measure`;
 - `cross_sections`;
 - `find_holes`;
 - `find_hole_patterns`;
@@ -151,10 +160,14 @@ The following operations become internal analysers used by `inspect_part`:
 - `find_bored_bosses`;
 - `find_countersinks`;
 - `locate_gate_defects`;
-- the relevant implementation from `feature_audit`.
+- the relevant implementation from `inspect_part`.
 
 They should remain independently tested Python functions. Consolidation applies to
 MCP exposure, not implementation structure.
+
+`measure` is deliberately excluded from this initial removal list. Its standalone MCP
+exposure can be reconsidered only after benchmark logs show that agents reliably use
+the in-`execute()` helper for the tight boolean-verification loop.
 
 `resolve` should become an `execute()` namespace helper unless usage evidence shows
 that agents need it as a standalone MCP call.
@@ -199,9 +212,8 @@ logs before changing exposure.
 
 ### 2. Finalise `inspect_part`
 
-Rename the issue-417 concept from `feature_audit` to `inspect_part`. Narrow its public
-arguments and expectation schema. Add automatic section-axis selection and bounded,
-compact output.
+Build on the issue-417 `inspect_part` checkpoint. Narrow its public arguments and
+expectation schema. Add automatic section-axis selection and bounded, compact output.
 
 ### 3. Reuse internal analysers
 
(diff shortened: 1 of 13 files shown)


pzfreo:
## Review of the tool-surface-consolidation proposal (re: this PR)

I reviewed the consolidation proposal against the current codebase (37 tools register by default in `server.py`), #417, and this PR. Short version: **the consolidation direction is sound, but it shouldn't block this PR.** Recommend decoupling.

### Ship this PR on its own
#417 asked for one audit helper; this PR delivers exactly that (`feature_audit`, +1085, ruff/mypy/tests green). The proposal's recommendation to *not* merge it and instead re-scope it as the entry point to a 12-tool surface redesign converts a shippable, well-tested increment into a multi-PR migration with a benchmark gate. That coupling is the main risk.

**Suggested for this PR — cheap, worth doing now:**
- Rename `feature_audit` → `inspect_part` before it goes public.
- Keep the `expected` schema narrow and fail-closed on unknown fields (already done here — good).

Everything else in the proposal (folding the recogniser cluster behind `inspect_part`, `--specialist-tools`, deprecating `--disable-tool-groups`, benchmark-gated default flip) is separate work now tracked in #421.

### Things #421 must resolve before any removal
1. **Drawing tools.** The proposal's target surface silently omits the whole `drawing` group (`render_drawing`, `inspect_drawing`, `lint_drawing`, `view_axes`, `save_drawing_annotations`, `suggest_view_layout`) — a real workflow with its own skill. "12–15 tools" isn't honest if it's partly achieved by dropping a capability. Needs an explicit decision.
2. **`measure` on the quick loop.** It's the most-used verify primitive ("call `measure()` after every boolean"). Folding it into a heavier audit call changes the tight `execute → measure` loop. The clean answer is probably "use the in-`execute()` Python helper for quick checks, `inspect_part` for checkpoints" — but that has to be stated, and reconciled with keeping `validate`/`compare` on the same "frequent, small contract" reasoning.
3. **`inspect_part` scope.** Compact default + 5 `detail` modes + a structured expectation schema is already trending toward the god-tool principle 6 forbids. Watch it.
4. **Sequencing.** Put the consolidated surface behind a flag and A/B it against the current default *before* flipping the registration default — the benchmark is a gate, not a postscript.

Tracking issue for the broader effort: #421.

pzfreo:
## Adversarial review

Reviewed `feature_audit.py` and its wiring (worker stub, subprocess dispatch, server tool) against the branch head, with two independent adversarial passes plus manual verification. **No crashes or session-killing paths found; validation is genuinely fail-closed.** Two findings worth addressing before this becomes the consolidated inspection entry point, plus two minor ones.

### 1. MEDIUM — a `PASS` can hide wrong-axis/wrong-bottom features when the expectation omits a qualifier

The tool description says *"A supplied feature category is an exact inventory … unexpected or ambiguously matched groups fail"* and the PR summary says *"exact one-to-one feature inventory matching."* The matcher is actually **many-to-one**: `_check_expected_groups` (feature_audit.py:151–166) sums `group["count"]` into `matched_counts` for every actual group that matches, and `_group_matches` only checks the keys the caller *supplied*. So two physically distinct actual groups collapse into one expected line whenever the expected line omits the qualifier that distinguishes them.

Concrete false PASS — exactly the defect class #417 cites ("holes placed on the wrong body surface"):
- Drawing: 4× Ø6 through-holes on the top face. Agent writes, from the dimension table: `{"holes":[{"count":4,"diameter":6}]}`.
- Defective model: 2× Ø6 on Z + 2× Ø6 on X.
- Trace: both groups match `expected[0]` on `diameter` only → neither "unexpected", neither "ambiguous" → `matched_counts[0] = 2 + 2 = 4` → `4 == 4` → **`status: "PASS"`**. Half the holes are on the wrong face.

The tool *can* catch this — but only if the agent also specifies `axis`. Omitting a qualifier silently weakens the verdict, which is precisely what the proposal's own expectation contract says it must not do. Note `test_feature_audit_rejects_unexpected_and_ambiguously_matched_groups` only exercises an extra group with a **different diameter** (matches zero expecteds → correctly flagged); it never covers an extra group that shares the specified qualifier, which is where the guarantee leaks.

Fix direction: track which actual groups are consumed and flag any expected line satisfied by >1 distinct actual group (or fail if any actual group is absorbed beyond a 1:1 map). At minimum, drop "exact one-to-one" from the copy and document that omitted qualifiers weaken the verdict.

### 2. MEDIUM (latent) — subprocess path-invariance is unproven for feature-rich parts

The tool's correctness rests on the in-process report equalling the subprocess report. But the bounded path re-serializes through STEP at ≥800 faces (`_bounded._FACE_GATE`), and the recognizers this tool leans on — hole *bottom* classification, boss height, pattern grouping — are more STEP-sensitive than `measure`'s volume/area (`measure.py` itself documents a cone's sign flipping on a STEP round-trip). The safety rationale in `_shape_op_subprocess.py` only justifies round-trip invariance for `measure`/`cross_sections`/`clearance`; `feature_audit` was added to the dispatch without extending that argument.

The only round-trip test (`test_feature_audit_round_trips_through_real_bounded_subprocess`) uses a plain `Box` — no holes, bosses, or patterns — so a divergence in exactly those recognizers would ship unnoticed, and it would hit precisely the large parts you can't eyeball. I could not *reproduce* a divergence (box+through/blind/counterbore/bolt-circle all matched in-process vs round-tripped), so this is a latent risk, not a demonstrated bug.

Fix direction: add a round-trip-equality test on a feature-rich part (blind hole + counterbore + bolt-circle) asserting the subprocess report equals the in-process report, and extend the safety comment to cover the new op.

### 3. LOW — duplicate/overlapping expected group lines are unsatisfiable

Two identical (or subset-overlapping) expected group lines make every matching actual group match >1 expectation → permanent "ambiguous" FAIL with no geometry that can pass, e.g. `{"holes":[{"count":2,"diameter":6},{"count":2,"diameter":6}]}`. Fail-closed (safe) but confusing; worth a clearer message or a validation-time reject.

### 4. LOW — malformed `expected` JSON raises a raw `JSONDecodeError`

`json.loads(expected)` at feature_audit.py:327 is unwrapped, so `expected='{bad'` surfaces a `JSONDecodeError`, whereas every other bad input (`'[]'`, `'null'`, `'42'`) gives a clean `ValueError("expected must be a JSON object")`. Surfaces as a tool error (doesn't kill the session) — cosmetic consistency only.

### Positives
Validation (`_validate_expectation`) is thorough and fail-closed — rejects bool-as-int, NaN/inf, unknown top-level and per-group keys, malformed vectors, tolerance-only no-ops. The `_pattern_record` refactor is a clean dedup and keeps the forward-compat handling for unknown pattern types. Div-by-zero in `_section_summary` is guarded; degenerate/2D shapes don't crash; param passthrough across server→worker→subprocess is correct.

Findings 1 and 2 matter most given #421's plan to make this the single inspection entry point — a checkpoint tool that can PASS a wrong-axis part, or diverge silently on large parts, undercuts the reason it exists.

pzfreo:
Addressed both review comments in `4fd2505`.

- Public API renamed to `inspect_part`; broader surface migration remains decoupled in #421.
- One expected line can no longer absorb multiple distinct actual groups. The wrong-axis `2×Z + 2×X` versus `4× Ø6` case now fails as underspecified.
- Duplicate and subset-overlapping expected lines are rejected at schema-validation time.
- Malformed JSON now returns a clean `ValueError`.
- The bounded round-trip test now compares complete inspection sections for a feature-rich part containing through, blind, counterbored, bolt-circle, and boss geometry.
- CI's missing worker-operation classification and MCP tool inventory were added.
- The proposal now explicitly retains drawing pending #421, preserves the quick in-`execute()` `measure()` loop, defers `detail` modes, and requires flag-based A/B validation before any default registration change.

Local verification: Ruff and mypy clean; full suite reached 1,039 passed / 1 skipped with one stale tool-list assertion, then the fixed assertion and all affected tests passed (40/40). Fresh CI is running.

pzfreo:
Addressed both review comments in `4fd2505`.

- Public API renamed to `inspect_part`; broader surface migration remains decoupled in #421.
- One expected line can no longer absorb multiple distinct actual groups. The wrong-axis `2×Z + 2×X` versus `4× Ø6` case now fails as underspecified.
- Duplicate and subset-overlapping expected lines are rejected at schema-validation time.
- Malformed JSON now returns a clean `ValueError`.
- The bounded round-trip test now compares complete inspection sections for a feature-rich part containing through, blind, counterbored, bolt-circle, and boss geometry.
- CI's missing worker-operation classification and MCP tool inventory were added.
- The proposal now explicitly retains drawing pending #421, preserves the quick in-`execute()` `measure()` loop, defers `detail` modes, and requires flag-based A/B validation before any default registration change.

Local verification: Ruff and mypy clean; full suite reached 1,039 passed / 1 skipped with one stale tool-list assertion, then the fixed assertion and all affected tests passed (40/40). Fresh CI is running.

PR #426, opened 2026-07-26
Handle missing face triangulation in renders

### LINKED ISSUE #425: render_view: handle faces with missing triangulation without NbNodes crash
## Background

Follow-up to the external contributor report on #398. The locator defects from that report were addressed by #423 and #424, but the related rendering failure remains.

On a mesh-derived solid with one or more faces for which `BRep_Tool.Triangulation_s(...)` returns `None`, `render_view()` can skip the entire shape and report:

```text
Skipped shapes (tessellation failed): 'NoneType' object has no attribute 'NbNodes'
```

The reported reproduction imported a 3MF, performed box subtractions, and healed with a tolerant fuse. The unmodified imported shape already contained the problematic faces, so the rendering failure does not require the boolean edit. `validate()` correctly reported base/refined untriangulated faces, and `locate_gate_defects()` now returns their coordinates, but rendering still dereferences the missing triangulation.

## Problem

A known per-face tessellation defect should not become an opaque whole-shape `AttributeError`. Rendering should either:

- render the triangulatable faces and report precisely which faces were omitted; or
- fail the shape cleanly with a structured, actionable missing-triangulation diagnostic.

The implementation must remain bounded. It must not add an unbounded retry or refinement ladder to the render worker.

## Acceptance criteria

- No `None.NbNodes` dereference when an OCC face has no triangulation.
- The render result identifies missing-triangulation failure explicitly rather than exposing a raw `AttributeError`.
- Prefer partial rendering of healthy faces when the rendering architecture supports it safely; otherwise preserve the current skipped-shape behavior with an actionable diagnostic.
- Existing render subprocess and operation timeout guarantees remain intact.
- Regression coverage includes a face whose triangulation lookup returns `None`.
- The behavior is documented in the changelog.

## References

- External reproduction: #398 comment by @beantownbytes
- Locator-side base-face fix and 3MF import: #423
- Ladder-aware locator fix: #424

pzfreo:
## Summary

- extract render meshes face-by-face after build123d's guarded mesh operation
- skip faces whose OCC triangulation is missing while rendering healthy faces
- report the affected 1-based face indexes and whether a partial render was produced
- preserve existing imported STL mesh behavior in both subprocess and in-process paths

## Root cause

`Shape.tessellate()` assumes every face has an OCC triangulation and dereferences the result without checking for `None`. A malformed or partially meshable shape therefore raised an opaque `None.NbNodes` error and caused the whole shape to be skipped.

## User impact

Partially meshable objects now render their healthy faces and return an actionable warning. Objects with no renderable faces fail cleanly with the affected face indexes instead of exposing an internal attribute error.

## Validation

- `uv run ruff format --check src/build123d_mcp/_tessellate_subprocess.py src/build123d_mcp/tools/render.py tests/test_render_tessellation_bounded.py`
- `uv run ruff check src/build123d_mcp/_tessellate_subprocess.py src/build123d_mcp/tools/render.py tests/test_render_tessellation_bounded.py`
- `uv run mypy src/build123d_mcp/_tessellate_subprocess.py src/build123d_mcp/tools/render.py`
- `uv run pytest -q` (`1061 passed, 1 skipped`)

Closes #425


pzfreo:
## Review

**What it does:** replaces `Shape.tessellate()` in the render path with a local face-by-face extraction (`tessellate_shape_faces`) that skips faces whose OCC triangulation is `None`, returns the 1-based indexes of those faces, and renders whatever is left. Fixes the raw `None.NbNodes` from #425.

**Verified locally:** `ruff check` / `ruff format --check` / `mypy src` clean; `tests/test_render_tessellation_bounded.py` 14 passed, and the render/view/drawing/export slice 296 passed, 1 skipped.

### The reimplementation is faithful — I checked

Reimplementing `tessellate()` is the risky part of this PR, so I diffed the output numerically against build123d's on Box, Cylinder, a boolean, a filleted solid, two compounds, a 5-face shell, and a bare face: **identical vertex counts, triangle counts, and signed volumes** in every case.

Worth noting the winding convention *is* different — build123d emits `(1,3,2)` for a `TopAbs_REVERSED` face, this emits `(2,1,3)`. Both are odd permutations, so normals point the same way; the matching positive signed volumes confirm it. Not a bug, just don't "fix" it later thinking it diverged.

Two incidental wins: it builds plain tuples instead of build123d `Vector` objects plus a second comprehension (cheaper), and the tuple triangles now match `_vtk_render_tesselated`'s documented `[(i,j,k), ...]` contract. CHANGELOG correctly targets the pending v0.3.80.

Also nice: the reported face indexes come from the same `TopTools_IndexedMapOfShape` 1-based ordering that `locate_gate_defects` uses for `mesh_untriangulated_face.face_index`, so "face(s) 3" in a render warning cross-references directly to the locator's output.

### Should fix

**1. Both fallback branches are no-ops, and their comments say otherwise.**

`if faces.Size() == 0:` claims it preserves STL mesh data that has no OCC faces. But build123d's `tessellate()` iterates `self.faces()` too — with zero faces it returns `([], [])`. I checked both halves: an STL import is a `Face` with **1** face carrying a triangulation, so it goes through the normal loop (which is what the PR's own STL test actually exercises); a STEP-round-tripped STL has **0** faces and old and new both return `(0, 0)`. The branch can only ever return empty, after paying an extra `mesh()` + `tessellate()`.

The `if not triangles and not missing_faces:` branch has the same problem: every face there has a non-`None` triangulation with zero triangles, and `tessellate()` reads those same triangulations. "build123d can still recover the original mesh" isn't supported by its implementation. (Strictly, the second `tessellate()` re-invokes `mesh()`, which could re-run `BRepMesh_IncrementalMesh` with identical parameters — but banking on a repeat remesh yielding different output is accidental, not designed.)

Neither branch has a test. I'd delete both.

**2. The record-the-result block is copy-pasted into two places.**

```python
if tris or not missing_faces:
    meshes[name] = (verts, tris)
if missing_faces:
    failed.append(_missing_faces_message(name, missing_faces, bool(tris)))
```

lives in both `_tessellate_subprocess.main()` and `render._tessellate_in_process()`. Only the in-process copy is tested — monkeypatch can't reach the subprocess, so the partial/all-missing behaviour of the *default* path is unverified. One shared helper fixes the drift risk and the coverage gap, and removes the cross-module import of the private `_missing_faces_message`.

### Minor

- `render.py:419` — the `from build123d_mcp._tessellate_subprocess import ...` sits *after* `meshes`/`failed` are declared inside the function. Move it above them.

### Out of scope, but the class isn't retired

The same unguarded `shape.tessellate()` remains at `export.py:46` (`_stl_write`) and `_shape_compare_subprocess.py:78` (`_tessellate_points`) — both would still raise raw `None.NbNodes` on the shape from #425. `validate.py:551` is inside a try/except, so it degrades to "fast check unavailable" rather than crashing. Not asking for a fix in this PR; flagging so it isn't assumed closed.

PR #433, opened 2026-08-07
Fold the session entry into v0.3.82

pzfreo:
`pyproject.toml` is at `0.3.82.dev0`, so the next release is **v0.3.82** — but the per-handle session work (#432) was filed under a `## v0.3.83` heading sitting above it.

Cutting v0.3.82 in that state would have produced release notes covering only the SDK v2 migration, silently omitting the session isolation that ships in the same release.

Both sets of changes are in the same release, so the session bullets move into v0.3.82's existing `### Added` section alongside the conformance entry. No content changed, only where it sits.

Docs only — no code touched.


pzfreo:
Closing as superseded: the HTTP session and session-tool release notes are already in the v0.3.82 section of CHANGELOG.md on main. This old branch now conflicts with main.

PR #440, opened 2026-08-08
Add native 3MF export format

zeekwan:
## Summary

- Adds `3mf` as a valid `export()` format alongside the existing `step`/`stl`/`dxf`/`svg`. Pairs naturally with the 3MF **import** support merged in #423 — export was the missing half.
- Implementation is stdlib-only (`zipfile` + `xml.etree.ElementTree`, no new dependency): a minimal core-spec 3MF (single mesh, no color/material extensions) intended to be readable by mainstream slicers (Bambu Studio, PrusaSlicer, Orca).
- Reuses the exact same `shape.tessellate()` call the existing STL writer already uses, so 3MF and STL vertex positions match exactly for the same shape.

## Motivation

STEP/STL/DXF/SVG cover CAD interop and 2D drawings, but 3MF is what Bambu Studio (and most modern slicers) prefer for direct print-ready import — it's mesh + build-item metadata in one file, avoiding some of the ambiguity STL has around units/orientation. Wanted this for a personal print workflow and figured it's generally useful given 3MF import already landed.

## Test plan

- [x] `uv run pytest tests/test_tools.py -k "3mf or export"` — 27 passed (covers the new `test_export_3mf` / `test_export_3mf_rejects_2d_shape`, plus existing export + the #423 3MF import tests, confirming no regressions against the already-merged import path). These tests validate zip/XML structure (valid archive, correct parts, well-formed `<vertex>`/`<triangle>` elements) — not a live open in slicer software.
- [x] `uv run ruff check` — clean
- [x] `uv run mypy src/build123d_mcp/tools/export.py` — clean
- [ ] Not yet confirmed opening the exported file in an actual slicer (Bambu Studio/PrusaSlicer/Orca) — happy to do that verification if useful before merge, just flagging it's not done yet.


pzfreo (commented):
Thanks for this, @zeekwan — genuinely nice contribution. Closing the export half of 3MF was on our list, and doing it stdlib-only with `zipfile` + `ElementTree` is exactly the right call; I appreciate you reusing the same `shape.tessellate()` the STL writer uses so the two formats stay consistent. Also thank you for being upfront in the test plan that the slicer check wasn't done — that flag is what pointed me straight at the one real issue.

**One blocker, and it's a one-liner.** `[Content_Types].xml` needs a `rels` default:

```xml
<Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/>
```

Without it the package isn't OPC-conformant, so lib3mf-based readers refuse it. I exported a Box from your branch and fed it back through this repo's own `import_cad_file`:

```
Lib3MFException 5: Content Types does not contain extension for relatioship.
```

Adding that single line to the identical file makes it import cleanly (volume 6000.0, 1 solid). So your instinct about the unticked checkbox was right — no slicer would have opened it.

**One test request.** Could you add a round-trip assertion — export, then `import_cad_file`, then check the volume? The current test verifies zip members and `<vertex>`/`<triangle>` structure, which is reasonable on its face but stays green on a file no reader accepts. A round trip would have caught the above, and it guards the format for whoever touches it next.

**Non-blocking, entirely optional.** Building one ElementTree element per vertex and triangle gets expensive on big meshes — I measured 128k triangles at 26.3 s / 142 MB for the XML stage versus 3.4 s for the streaming STL writer. Around 500k triangles that crosses our 120 s export budget and the worker gets killed, taking the session with it. Streaming the XML as text would fix it, but I'm happy to take that as a separate PR rather than grow this one.

Two tiny things while you're in there, if you like: the 3D-shape rejection message still reads "Use 'step' or 'stl'", and a `CHANGELOG.md` entry would be good since we keep that current.

Fix the content types and add the round trip and I'll merge. Thanks again for picking this up.

zeekwan:
Appreciate the review! Fixed both, pushed to the branch.

**Content types**: added the missing `<Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/>`. To make sure this doesn't regress silently, I reproduced your finding first — reverted the fix locally and ran the new round-trip test against it, got the identical error:

```
lib3mf.Lib3MF.ELib3MFException: ... Content Types does not contain extension for relatioship.
```

Restored the fix, test passes, `import_cad_file` round-trips a `Box(10,10,10)` back to volume≈1000, 1 solid.

**Round trip test**: added `test_export_3mf_round_trip` — export → `import_cad_file` → assert volume/solids, right next to the existing structural test rather than replacing it. Confirmed it's the one that would've caught this: it fails on the unfixed code with your exact `Lib3MFException 5`, passes once the content-types fix is in.

**Two tiny things**: fixed the stale "Use 'step' or 'stl'" message (now includes 3mf), added a CHANGELOG entry.

**Streaming XML perf**: agreed, leaving that out of this PR as you suggested.

All 28 export/3mf-related tests pass, plus `ruff check`, `ruff format --check`, and `mypy` clean on the touched files.

pzfreo (approved):
Thank you, @zeekwan — this is exactly how you'd want a review round to go.

I re-verified the fix rather than just reading it: I reverted only the `<Default Extension="rels" .../>` element on your branch and re-ran the 3MF tests. `test_export_3mf_round_trip` failed with the original `Lib3MFException 5`, while the other seven — including the pre-existing structural test — still passed on the broken file. So the test you added is precisely the one carrying the weight, and it fails for the right reason. Reproducing the bug first before fixing it was the right instinct, and it's why I trust this.

CI is green across the full matrix (ubuntu/macOS/Windows × Python 3.12–3.14 × build123d 0.10/0.11, plus conformance), and `ruff` and `mypy` are clean. The CHANGELOG entry and the corrected format-list message are both good.

Merging. The XML streaming work is genuinely a separate concern and you were right to leave it out — I'll open a follow-up issue for it, and if you'd like to take that one too you'd be very welcome.

Thanks again for filling in the missing half of 3MF support, and for the care in the write-up — flagging your own unverified checkbox is what made this quick to review.

PR #451, opened 2026-08-25
Add mesh_holes and mesh_section: cross-section analysis for meshes

dcox79:
## Why

`import_cad_file` on an STL yields a single `Face` — a shell with no solids and no topology — so `find_holes`, `find_countersinks` and the rest of the recogniser family return nothing on a downloaded mesh. The tool's own docstring already notes *"STL imports produce a shell (volume=0) rather than a solid."*

That is the common case when remixing a model from Printables or Thingiverse: you need the existing mounting features in someone else's STL before you can design a part that bolts to it. Today there is no way to get them from this server.

## What

Two read-only tools in `tools/mesh_analysis.py`:

- **`mesh_holes`** — slices all three axes and reports `{axis, diameter, location, span, depth, through}` per bore.
- **`mesh_section`** — loops on one plane, each with an `enclosed` flag, plus a count of enclosed passages.

Both go through `Shape.tessellate`, so they work on solids too — useful as a second opinion on `find_holes`, and as the only way to prove an internal duct is enclosed rather than an open groove, which no render can show.

## Failure modes it had to survive

Each of these produced a confidently wrong answer during development and is now a regression test:

1. **Opposed blind pockets merged into one through hole.** Two pockets bored into opposite faces share a cross-plane centre and a diameter; keyed together they collapse into one record spanning the full width. Records are emitted per contiguous run, not per key.
2. **A pocket shallower than one sample step vanished.** A bar with an identical pocket in each end read as having only one. Bores open on faces, so the sampler now sweeps finely just inside each face.
3. **Loop count is not containment.** A groove cut clean across a bar splits the section into two disjoint outlines; counting `loop_count - 1` calls an open groove an enclosed passage — exactly the distinction these tools exist to make. Containment is ray-cast per loop.
4. **A bore fragmented into four records** where tessellation chained badly. Gaps narrower than the bore are dropouts, not material, and are merged.

## Verification

Against a real downloaded model (Snapmaker U1 5x build plate holder, 2204 triangles): 8 bores per rail and 8 per frame, with depths that separate a 4 mm heat-set insert pocket from a 6.2 mm clearance hole and a 3 mm counterbore. `find_holes` reports 0 on the same import.

`tests/test_mesh_analysis.py` uses self-contained fixtures only — no third-party geometry.

## Wiring

All four steps from `CLAUDE.md`: tool function → `WorkerSession` stub → `server.py` registration → classified in `test_worker_boundary_coverage.py`.

## Test run

`1114 passed, 33 skipped, 1 failed`. The one failure is `test_export_symlink_escape_rejected`, which fails identically on unmodified `main` on this machine — a Windows symlink-privilege issue, unrelated to this change.

https://claude.ai/code/session_01M9AUHzL9uAUCzboxbBjWn3


pzfreo:
Really nice piece of work — this fills a genuine gap, and the failure-modes section in the description is the kind of thing I wish every PR had. The approach is sound. A few things to fix before merge; I reproduced each with probe scripts against this branch.

## Blocking

**1. `_enclosed_flags` uses `any()` instead of odd nesting depth** (`tools/mesh_analysis.py:125`)

Ray-casting per loop was the right call, but containment needs a parity count — otherwise solid material inside a cavity reads as a bore:

```python
Box(60, 60, 40) - Cylinder(20, 40) + Cylinder(3, 40)
# the solid Ø6 post comes back as {diameter: 6.0, depth: 39.965, through: true}
```

Count how many polygons contain the probe, take `count % 2 == 1`.

**2. `_merge_spans` re-merges opposed pockets when the land is thin** (`:323`)

The comment says *"the land between them is far wider than they are"* — true of the 40 mm test bar, not in general. Two Ø6 × 3 mm pockets in a 10 mm plate report as one `{depth: 9.991, through: true}`. That's heat-set inserts from both faces, a very common remix case. Dropout gaps scale with the sample step, so the threshold probably wants to key off `step` rather than `diameter`.

## Should fix

**3. `_chain` drops one vertex per loop** (`:82`) — `loop` starts empty and the first append is `start`'s *neighbour*, so `start`'s own point never lands in the output. A 4-segment square returns a 3-point triangle; genuine 3-segment loops vanish at `len(loop) >= 3`.

**4. `_chain`'s weld key splits coincident endpoints** (`:63`) — `round(p / weld)` puts identical endpoints in different cells near a `.5` boundary. On an L-bracket, `y = -24.6875` keys to both `-24688` and `-24687`; the outline fragments and a 10 mm through hole reports `depth: 8.681`. Falling back to neighbour cells on a miss would cover it.

**5. `through` compares to the global bbox extent** (`:295`) — a hole clean through a 10 mm flange on a 110 mm bracket reports `through: false`. Wants to be relative to the local wall, not the whole part.

**6. `_MAX_TRIANGLES = 400_000` outruns the 120 s exec budget** (`:21`) — 275 k triangles measured at 372 s, well under the cap and an ordinary STL size. The timeout SIGKILLs the worker, and `worker.py:436` doesn't replay `import_cad_file`, so the imported model is lost with no way back. ~50 k completes comfortably.

## Minor

- `:253` — `slices=0` / `weld=0.0` raise a bare `ZeroDivisionError`, `slices=-5` silently returns `count: 0`. `axis` is validated; these should be too.
- `server.py:556` — docstring describes `enclosed_passages` as "counts loops after the largest", the `loop_count - 1` heuristic your own tests call wrong. Also omits `enclosed` from the loop schema.

## Tests

Every fixture is a 40 mm bar with no solid islands, which is why 1 and 2 slip through. Worth adding a bore-inside-a-cavity case, opposed pockets in a plate thinner than the bore, a through hole in a short flange, and a unit test on `_chain`'s vertex count.

Ping me when these are in and I'll take another pass — keen to get this landed.

pzfreo:
@dcox79 Thanks again for contributing this. I pushed a maintainer follow-up to your branch and brought it up to date with `main`.

The follow-up addresses the review findings: nesting parity for solid islands, complete loop vertices and endpoint welding, separate opposed pockets even with a thin land, local wall thickness for `through`, input validation, and a lower triangle cap to stay within the worker budget. I added regression cases for those geometries and updated the MCP tool inventory test.

Locally, the mesh and worker tests passed (55 tests), as did type, lint, and format checks. The full suite had one tool inventory expectation failure; I fixed it and reran that test successfully. The [GitHub CI matrix](https://github.com/pzfreo/build123d-mcp/actions/runs/36032422421) is running now.

If you still have the downloaded STL from your original check, I'd welcome a recheck on this branch, especially of the reported bore depths.

pzfreo (commented):
Hi @dcox79, thanks for this contribution. Cross-section analysis for imported meshes fills a real gap, since `find_holes` returns nothing for an STL, and the notes on failure modes (opposed blind pockets, thin end walls) show a lot of careful thought.

I'm Claude, reviewing on Paul's behalf. I ran the tool on a few synthetic cases and found four things worth looking at, detailed in the inline comments. The first two are the most important: one concerns the worker time budget, the other the triangle cap on real downloaded STLs. The other two are edge cases in hole classification.

Happy to be corrected on any of these if I've misread the intent.

— Claude

src/build123d_mcp/worker.py
```diff
@@ -979,6 +979,30 @@ def clearance(self, object_a: str, object_b: str) -> str:
...
+        self,
+        object_name: str = "",
+        axis: str = "Z",
+        position: float = 0.0,
+        tolerance: float = 0.1,
+        weld: float = 0.001,
+    ) -> str:
+        raise NotImplementedError
+
+    @_op(_tool(f"{_T}.mesh_analysis:mesh_holes"), _export_budget)
```
pzfreo: `mesh_holes` runs under `_export_budget` (60 s by default), but its run time grows roughly with triangle count × (~71 sample planes per axis + refine slices per candidate + 9 ray casts per span), all in pure Python. In my tests:

- plate with 9 holes, ~22.7k triangles: **19 s**
- plate with 16 holes, ~47.4k triangles (still under the 50k cap): **130 s**

The second case overruns the budget, the worker is killed, and the user loses their session state. `slices` also has no upper limit, which makes this worse.

Some possible mitigations: base the work limit on triangles × planes rather than triangles alone, cap `slices`, check a deadline inside the sweep and return a partial result marked as such, or vectorise the slicing with numpy (already on the allowlist). Any of these would help. The main aim is that an input which passes validation can't exceed the op budget.

— Claude

src/build123d_mcp/tools/mesh_analysis.py
```diff
@@ -0,0 +1,476 @@
...
+
+
+def _triangles(shape, tolerance: float):
+    if not math.isfinite(tolerance) or tolerance <= 0:
+        raise ValueError("tolerance must be a positive finite number")
+    verts, tris = shape.tessellate(tolerance)
+    if len(tris) > _MAX_TRIANGLES:
+        raise ValueError(
+            f"mesh has {len(tris)} triangles (limit {_MAX_TRIANGLES}); "
+            "raise `tolerance` to tessellate more coarsely"
```
pzfreo: A problem with this advice: for a Face from `import_stl`, `tessellate()` returns the triangles stored in the file whatever the tolerance. I checked with a 98,864-triangle STL and got 98,864 triangles at tolerance 0.1, 1.0 and 10.0.

Downloaded models from Printables and Thingiverse (the main use case here) are often well over 50k triangles. For those the tool always refuses, and raising `tolerance` never helps. Perhaps the message could say the mesh needs decimating before import? The cap itself might also be worth revisiting together with the performance point on `worker.py`.

— Claude

src/build123d_mcp/tools/mesh_analysis.py
```diff
@@ -0,0 +1,476 @@
...
+        (diagonal, -diagonal),
+        (-diagonal, diagonal),
+        (-diagonal, -diagonal),
+    ]
+    spans = []
+    for du, dv in offsets:
+        hits = _axis_intersections(tris, axis, cu + du, cv + dv)
+        if len(hits) < 2:
+            continue
+        wall_start, wall_end = hits[0], hits[-1]
```
pzfreo: `hits[0]` / `hits[-1]` take the first and last intersections along the whole ray, not the wall that surrounds the bore. So any material further along the drilling axis makes a real through hole read as blind.

I tested a C-clip: a 3 mm bottom flange with a Ø3.4 through hole, and a top flange above it across an air gap. The result was `span [0.017, 2.943]`, `through: false`. U-channels, clips and overhangs would all hit this.

One fix would be to pick the pair of consecutive hits that brackets the detected run (`start`/`end`) instead of the outermost pair.

— Claude

src/build123d_mcp/tools/mesh_analysis.py
```diff
@@ -0,0 +1,476 @@
...
+    insert sits in. A hole shows up as a round loop only on slices normal to
+    its own axis, so scanning a single axis finds a fraction of the part.
+
+    This is the mesh counterpart to find_holes(), which needs real topology and
+    returns nothing for an imported STL. It reports what the cross-sections
+    show and does not classify counterbores, countersinks or thread forms.
+    """
+    from build123d_mcp.tools.measure import _resolve_shape
+
+    if slices < 1:
```
pzfreo: A small edge case: `slices=1` passes validation, but then `step` equals the full extent and there are no uniform planes, only the face-adjacent sweeps. `_runs` joins positions closer than `1.5 × step`, so opposed blind pockets at the two ends merge into one run. `_refine_span` then extends it face to face, and you get the "two pockets reported as one through hole" failure the docstring guards against. `_runs` doesn't have the barrier check that `_merge_spans` has.

This only happens at `slices=1`, so requiring `slices >= 2` (or some larger minimum) is probably enough.

— Claude

pzfreo:
Thanks for the second pass. I reproduced the C-clip and `slices=1` cases and pushed fixes in `b0c1f0d`. Runs now split at pocket-floor crossings, and the local-wall check chooses the material interval surrounding the detected bore.

For the scale issues, slicing and ray checks now use NumPy. The triangle cap is 200k, and the error advises decimating an imported STL; changing tessellation tolerance cannot simplify triangles already stored in an STL. A fixed 130,240-triangle mesh with 16 holes completed locally in about 19 seconds. The scan also checks a deadline ahead of the parent watchdog so it can return an error while preserving the worker session.

One detail on the budget: `_export_budget` has a 60 s floor, but the default session execution timeout makes its default 120 s. The reported 130 s case still exceeds it.

The focused tests and static checks pass locally. [CI for this commit](https://github.com/pzfreo/build123d-mcp/actions/runs/36050303557) is running.

Issue #465, opened 2026-08-28
Deprecate the engineering-drawing tool surface and point users at draftwright


pzfreo:
Drawing generation has a maintained home in [draftwright](https://github.com/pzfreo/draftwright), which publishes its own agent skill. This server should stop carrying a second implementation and point users there.

This follows the same split already applied upstream — `build123d-drafting-helpers` ADR 0007, *"helpers renders; draftwright reasons"* — and the recognition migration tracked in #464.

## Why now

- **Drawing is the least reliable surface here.** Three of the open anchoring bugs live in it, all the same root cause (a centroid anchor paired with origin-relative expectations).
- **Part of it is already deprecated upstream.** Helpers 0.13.0 deprecates `lint_drawing`, `find_overlaps`, `find_interferences` and `LintIssue` — frozen, and to be removed. `tools/lint_drawing.py` is on borrowed time regardless of this decision, and this resolves the "what about lint?" half of #464.
- **Duplication invites disagreement.** Two drawing implementations against one geometry kernel is the failure mode #453, #454, #456 and #462 were all instances of.
- **Tool-surface pressure.** 7 of 53 registered tools, which bears on #421.

## Surface to deprecate

**Tools (7)** — `render_drawing`, `inspect_drawing`, `lint_drawing`, `save_drawing_annotations`, `suggest_view_layout`, `view_axes`, and the `build123d_drafting_api` accessor.

**Modules (~2,000 LOC)**

| file | lines |
|---|---|
| `drafting_cookbook.py` | 699 |
| `tools/suggest_view_layout.py` | 378 |
| `tools/lint_drawing.py` | 328 |
| `tools/inspect_drawing.py` | 255 |
| `tools/view_axes.py` | 134 |
| `tools/render_drawing.py` | 87 |
| `tools/drafting_api.py` | 81 |
| `tools/save_drawing_annotations.py` | 44 |

**Skill** — `skills/b123d-drawing/SKILL.md` (689 lines) and the `drawing` entry in `install_skill()`.

**Resources** — `build123d://drafting`, `build123d://drafting-api`, `build123d://skill/drawing`.

## Explicitly staying

- **`render_view` in every format, DXF included.** That is model rendering, not drawing — an HLR projection of the live solid, useful for downstream 2D CAD regardless of who generates drawings. #455 and #435 remain valid bugs against it.
- Modeling / edit / repair skills, `measure`, `validate`, `export`, the recogniser tools.
[...]

Issue #457, opened 2026-08-26
suggest_view_layout() pairs a centroid look_at with origin-anchored cameras, producing tilted elevations


mironoprea:
## Summary

`suggest_view_layout()` returns a `look_at` at the part **centroid** but pairs it with fixed camera positions anchored at the **origin**. For any part whose centroid is not at the origin, the camera-to-look_at axis is tilted, so the front and side views come out as slightly rotated projections rather than true orthographic elevations.

The plan view is unaffected, which is what makes this hard to notice.

## Reproduction

A 60 x 40 x 10 mm plate sitting on z=0, so centroid `(0, 0, 5)`:

```
suggest_view_layout()
-> front: camera (0, -100, 0), look_at [0, 0, 5]
   side:  camera (100, 0, 0),  look_at [0, 0, 5]
   plan:  camera (0, 0, 100),  look_at [0, 0, 5]
```

Projecting with those values:

```
front view   projected 60.000 x 11.985    for a 10 mm-thick part
side view    projected 40.000 x 12.984    for a 10 mm-thick part
```

Confirmed analytically. The view direction is tilted by `atan(5/100) = 2.862 deg`, and:

```
40*sin(2.862) + 10*cos(2.862) = 11.985    (matches front, to 3 dp)
60*sin(2.862) + 10*cos(2.862) = 12.984    (matches side,  to 3 dp)
```

Offsetting each camera by `look_at` corrects it exactly:

```
front (0, -100, 5)  ->  60.0000 x 10.0000
side  (100, 0, 5)   ->  40.0000 x 10.0000     (residual 2e-7, HLR tessellation)
plan  (0, 0, 105)   ->  unchanged
```

## Source

`tools/suggest_view_layout.py:25-28` -- camera positions are fixed constants through the origin:

```python
"front": {"camera": (0, -100, 0), "up": (0, 0, 1)},
"plan":  {"camera": (0, 0, 100),  "up": (0, 1, 0)},
"side":  {"camera": (100, 0, 0),  "up": (0, 0, 1)},
"iso":   {"camera": (80, 80, 80), "up": (0, 0, 1)},
```

`tools/suggest_view_layout.py:344` -- but `look_at` is set to the centroid:

```python
look_at = [round(cx, 4), round(cy, 4), round(cz, 4)]
```

The two are never reconciled. The docstring at `:262` states the intended invariant explicitly:

> The bounding-box half-extents map directly to page extents when look_at is at the part centroid

[...]

PR #466, opened 2026-08-28
Anchor render_view DXF/SVG projections at the world origin (#455)

### LINKED ISSUE #435: render_view: no bottom/back view preset, and elevation offsets are unpredictable
## Summary

`render_view` exposes `direction`: `top`, `front`, `side`, `iso`. There is no way to look at a part's **underside**. For anything designed to be printed face-down — a phone case back, a bracket's mounting face, any shelled part — the underside is the interesting side, and it is the one view you cannot ask for.

The available workaround is to guess an `elevation` offset, but the offsets are not predictable because the docstring says only:

> `azimuth/elevation: camera rotation in degrees applied after the direction preset`

It never states what each preset's **baseline** azimuth/elevation is, so you cannot compute the offset you need — only guess and re-render.

## Concrete cost

Designing a phone case, trying to see the back:

| Call | Result |
|---|---|
| `direction="iso", elevation=-35` | edge-on sliver, part invisible |
| `direction="iso", elevation=-105` | correct back view |

Two of seven renders in the session were wasted this way. Each render is a multi-second VTK round trip, and for an LLM caller each wasted image is also wasted context.

## Suggested fix

1. Add a `bottom` preset (and ideally `back` / `left` for completeness). `top` already exists, so `bottom` is the obvious gap.
2. Document each preset's baseline azimuth/elevation in the tool description, so `elevation=` offsets are computable rather than guesswork.

Either one alone would have avoided the wasted renders; (1) is the smaller change.

## Version

build123d-mcp 0.3.82.dev372, build123d 0.11.1, macOS.

### LINKED ISSUE #455: DXF export writes centroid-relative coordinates, not model coordinates
## Summary

`render_view(format="dxf")` writes coordinates that are **relative to the part centroid**, not to the model origin. The shape is exact; the origin is displaced. Importing the DXF into CAD expecting model coordinates puts every feature off by the centroid, and the error grows with how asymmetric the part is.

## Reproduction

Test part: a 180 x 100 x 12 mm plate with two dia-20 bosses, a bolt circle, a counterbore, a countersink, a blind hole and a tap-drill bore -- deliberately asymmetric.

```
render_view(direction="top", format="dxf", save_to="testpart_top.dxf")
```

Parsing the resulting DXF:

```
reported bbox   x -90.3297 .. 89.6703     y -50.6787 .. 49.3213
expected bbox   x -90.0000 .. 90.0000     y -50.0000 .. 50.0000
offset          (-0.3297, +0.6787) on EVERY entity, identically
```

The part's centre of mass is `(0.3297, -0.6787, 6.2012)`. The offset equals the **negated centre-of-mass xy**, matched to 5 decimal places on both axes:

```
dxf_xy = model_xy - centre_of_mass_xy
```

The geometry itself is correct -- every radius exact to 4 dp (`r=10.0000` both bosses, `6.0000` countersink major, `5.5000` counterbore, `4.0000` blind + boss bore, `3.4000` tap, `3.0000` both dia-6 drills, `2.5000` x6 bolt circle), and the overall spans are exactly `180.0000 x 100.0000`. Only the origin moves.

## Source

`tools/render.py:646` -- the projection is anchored on an aggregate of `shape.center()` rather than on the world origin:

```python
# Aggregate bounding centre across all shapes for look_at
centres = [shape.center() for _name, shape, _c in shapes]
centre = sum(centres, Vector(0, 0, 0)) * (1.0 / len(centres))
```

## Why this matters more than the magnitude suggests

The offset here is small (0.33, 0.68 mm), but it is:

- **silent** -- nothing in the output indicates the coordinates are relative,
- **part-dependent** -- it scales with asymmetry, so it differs per part and cannot be corrected once and forgotten,
- **invisible to inspection** -- the shape and all dimensions are correct, so the file looks right until it is placed against other geometry.

For a format whose purpose is fabrication handoff and downstream 2D CAD, model-absolute coordinates are the expected contract.

## Suggested fix

Anchor the DXF projection at the world origin. If the centroid anchor is deliberate for framing reasons, the offset should at minimum be returned alongside the file (and ideally written as a DXF `$INSBASE`, or noted in the tool description) so a consumer can compensate.

## Unrelated doc note, in the tool's favour

`render_view`'s description says DXF is emitted as *"HLR projection as parseable polylines"*. It does not emit polylines -- it emits **29 true `CIRCLE` entities and 8 `LINE` entities**. Arcs are exact rather than tessellated, which is better than documented. Worth correcting the description.

## Environment

- build123d-mcp 0.3.82, build123d 0.11.1, Python 3.12, Windows 11, `BUILD123D_IN_PROCESS=1`

Found in the same testing pass as #453.

pzfreo:
Fixes #455. Also fixes #435.

## The defect

`project_to_viewport` returns 2D coordinates relative to its `look_at` point, and `_viewport_origin_for` (`render.py:646`) set that point to an aggregate of `shape.center()`. Every coordinate written to DXF was therefore displaced by the part's centre of mass.

Reproduced exactly as reported — the offset equals the negated centre-of-mass xy, on every entity, to the last decimal.

The shape was *exact*: every radius and span correct. Only the origin moved. That is what makes it worth fixing rather than documenting:

- **silent** — nothing in the output says the coordinates are relative;
- **part-dependent** — the offset scales with asymmetry, so it cannot be corrected once and forgotten;
- **invisible to inspection** — the file looks right until it is placed against other geometry.

For a format whose purpose is fabrication handoff, model-absolute coordinates are the contract.

## The fix

Anchor `look_at` at the world origin and offset the camera along the view direction.

**This is safe because the HLR projection is parallel** — I checked rather than assumed, since a perspective projection would have made this change alter the drawing. Moving the camera along the view direction changes only where the 2D origin sits, never the projected shape:

| test | result |
|---|---|
| part 1e6 mm from the origin | spans exact to 6dp |
| camera distance ×10 → ×100,000 | spans identical |
| centroid vs origin anchor | identical spans, identical edge counts |

After the fix, on a deliberately asymmetric plate:

```
model bbox   x -90.0..90.0     y -50.0..50.0
DXF bbox     x -90.0000..90.0000   y -50.0000..50.0000
offset       dx=+0.000000  dy=+0.000000
circle centres  {(-55.0, 25.0), (-40.0, -30.0), (20.0, -35.0), (60.0, 20.0)}   ← true model positions
radii           [3.0, 4.0, 10.0]                                                ← unchanged
```

## Scope

**SVG shares the helper** and gets the same fix. `ExportSVG` auto-fits with a margin, so its appearance is unchanged while anyone parsing it gains model coordinates.

**The VTK/PNG path is untouched** — it has its own camera and never called this helper.

**This also fixes #435**, whose unpredictable elevation offsets are the same root cause: a centroid anchor paired with origin-relative expectations.

## Documentation correction

The reporter noticed the DXF is emitted as **true `CIRCLE` and `LINE` entities**, not the "parseable polylines" the docs claimed — arcs are exact rather than tessellated, which is better than documented. Corrected in all three places it was stated: `server.py`, `render.py`, `llms.md`. There's a test asserting no `LWPOLYLINE`/`POLYLINE` is emitted, so the claim stays true.

## Verification

`tests/test_render_dxf_coordinates.py` is new — 10 tests covering the reported top-view case, circle centres at true model positions, a guard that the offset no longer tracks the centroid, **all three cardinal views** (each projects a different pair of model axes, and the issue only exercised plan), geometry unchanged, a part far from the origin, the arc-entity claim, and SVG still rendering.

Full gate: mypy clean, ruff clean, **1183 tests passed**.

## Note on #457

The sibling report #457 (`suggest_view_layout` pairing a centroid `look_at` with origin-anchored cameras) is the same class of bug but is **not** fixed here — that tool is part of the drawing surface proposed for deprecation in #465. `render_view` stays either way, which is why this one is worth fixing now.

https://claude.ai/code/session_01AFuVqG4wK52MzEjHqGX93r


pzfreo:
Correction: this PR's description claimed it also fixed #435. It does not — I misread that issue as sharing #455's centroid-anchoring root cause. #435 is about missing `bottom`/`back` view presets and undocumented baseline azimuth/elevation for the existing ones, neither of which this touches. #435 is reopened.

